PR #15, opened 2011-08-05
Add mouse support and a few changes/fix

livibetter:
I know I should have asked when I started adding mouse support. So, I will highlight the changes I have made:

**mouse support**

 * add
  * `class double_press_input_filter` for double click,
  * `def listen_mouse_event` like `def listen`, and
  * `def mouse_event` like `def keypress` for mouse support.
 * change some event handlers because the argument difference of key event and mouse event, e.g.
    `def move_down(w, size, key)` to `move_down(w, size, *args)`. Almost all of those handlers I changed don't use `key` at all.
 * add some mouse event handler, which are not documented in HELP yet.

**changes**

 * change `change_rhs_box` sizing method, which cause both above and below the selected sidebox resize, which is strange to me. I changed it so it would only affect the box below.
 * add `non_post_mortem_only` decorator,  just try to reduce few lines.

**fix**

 * exception when key press on empty variables list.

It's fine if no one else wants the mouse support; however if it's accepted, but some naming or style/method don't fit into pudb, please let me know I will add additional changes to meet the need.

The two **changes** I made, especially first one, might need more input because not everyone likes the sizing method I like as I don't like the current.


livibetter:
More than 3 years, closing to clean up my PR.

asmeurer:
I still would like to see this implemented by the way.

inducer:
Now that I know about Shift+click, I'm much more in favor of this. (Thanks @denilsonsa!) @asmeurer - do you have time to fish out the mouse specific parts of this PR?

asmeurer:
Not right now but I'll probably eventually do it (if someone else doesn't first). 

Also, Shift+click is specific to whatever terminal you are using. In iTerm on Mac OS X, it's Option-click that does the same thing. 

And regardless, we should probably add a setting to disable it if it is implemented.

PR #90, opened 2013-10-31
A way to un-breakpoint a set_trace() line

asmeurer:
This sounds kind of odd, let me explain.  I usually enter the debugger with an explicit call to set_trace().  Often, once I've hit that line a few times, I no longer need to stop there, but I don't want to restart the process.

Just as I can manually turn a regular line into a breakpoint with the 'b' command, I would love a way to turn a set_trace() line back into a regular, non-breaking line.  The set_trace function would still be called of course, but could examine the calling line to see if it had been marked as non-breaking, and continue.

### DIFF
--- a/pudb/__init__.py
+++ b/pudb/__init__.py
@@ -148,7 +148,7 @@ def set_trace():
 
 def _interrupt_handler(signum, frame):
     from pudb import _get_debugger
-    _get_debugger().set_trace(frame)
+    _get_debugger().set_trace(frame, as_breakpoint=False)
 
 
 def set_interrupt_handler(interrupt_signal=DEFAULT_SIGNAL):
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -160,16 +160,24 @@ class Debugger(bdb.Bdb):
         This is exactly the same as Bdb.set_trace(), sans the self.reset() call.
         """
         if frame is None:
-            frame = sys._getframe().f_back
+            frame = thisframe = sys._getframe().f_back
+        else:
+            thisframe = frame
         # See pudb issue #52. If this works well enough we should upstream to
         # stdlib bdb.py.
         #self.reset()
+
         while frame:
             frame.f_trace = self.trace_dispatch
             self.botframe = frame
             frame = frame.f_back
-        self.set_step()
-        sys.settrace(self.trace_dispatch)
+
+        if (thisframe, thisframe.f_lineno) not in self.set_traces:
+            self.set_step()
+            sys.settrace(self.trace_dispatch)
+            self.set_traces[(thisframe, thisframe.f_lineno)] = False
+        else:
+            self.set_continue()
 
     def save_breakpoints(self):
         from pudb.settings import save_breakpoints
@@ -189,6 +197,7 @@ class Debugger(bdb.Bdb):
         self._wait_for_mainpyfile = False
         self.current_bp = None
         self.post_mortem = False
+        self.set_traces = {}
 
     def restart(self):
         from linecache import checkcache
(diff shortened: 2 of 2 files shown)


inducer:
If I Ctrl-C in the middle of execution with your patch applied, there's a spurious (and confusing) breakpoint shown in the source. (Easy to reproduce with `debug_me.py` in the source.)

asmeurer:
That was easy enough to fix, but there seems to be another issue that comes up when there is a normal breakpoint in the same file. I'm still working on that.

asmeurer:
OK, I fixed that bug, again on an airplane (this time with some pretty hard-core non-PuDB debugging). Lesson learned: be careful with `+=`. If you use it on on a list, and that list is referenced somewhere else, it will screw up the other code, and is ***very*** hard to track down.

inducer:
In b979fc59, something is still amiss. Here's a reproducer:

1. Add a `pu.db` in line 33of `debug_me.py`.
1. Run `./try-the-debugger.sh`
1. Hit `n` 4 times
1. Hit `c`.
1. When the `pu.db` is hit, a spurious breakpoint appears on line 36.

inducer:
Forgot to say: Thanks for your work on this!

asmeurer:
Doesn't reproduce for me. Does it actually break at the spurious breakpoint?

asmeurer:
By the way, I noticed that the fake breakpoints are a little off because normal breakpoints stop on the line itself, but set_trace stops on the line after. But I think the fix here is to actually go in and make the set_trace breakpoints color differently.

asmeurer:
Oh I think I might be seeing what you are seeing now. It only seems to happen in Python 2. Except for me the spurious breakpoint is on line 34 (right below the `pu.db`).

asmeurer:
Actually, I think the thing I described is the source of the problem. It thinks the line number is the line after the set_trace (which could be any line depending on the flow of the code). Any thoughts on how to fix this?

asmeurer:
And why is it Python 2 only? Did Python 3 change the behavior of `frame.f_lineno`?

asmeurer:
Searching around, I found some vaguely similar things in the Python issue tracker, like http://bugs.python.org/issue17697#msg186637. But I haven't dug deep enough to know if they are the same, or how to fix this issue. Again, if you have any thoughts, I'd love to hear them.

asmeurer:
Hmm, now I can only seem to reproduce it in Python 3.

asmeurer:
OK, I have applied the patch from https://bugs.python.org/issue16482, which was mentioned in https://bugs.python.org/issue17697 to fix the problem, and it seems to work (and based on my understanding of the problem, it seems like it should work). It looks like this is a workaround for a bug in the interpreter itself, where line numbers can be reported incorrectly on frames with traces set on them when the global trace is removed.  The workaround, as I understand it, is to remove the traces when continuing so that the line numbers get reported correctly.  As such this probably deserves some testing. Everything seems to work in my limited testing. I can definitely reproduce your issue and this latest commit fixes it. 

Also see https://bitbucket.org/xdegaye/pdb-clone/commits/123d1b6db6649f4b54712db321865fda55395488, which looks like it goes even further than my patch to toward working around this issue (see also the [current code](https://bitbucket.org/xdegaye/pdb-clone/src/6fa9eb68fe5d767d24af3d319c93f9bb9a3ad61b/lib/pdb_clone/bdb.py?at=default#cl-691)). I'm unclear why this fix would be necessary for anything other than `continue`, however, and it seems to me that the worst that can happen is that another behavior might lead to an incorrect line number being reported somewhere. 

At any rate, please test this. To my knowledge, it should be ready to go (again).

inducer:
In my limited testing, this seems to do the trick. This whole 'disable set_trace' thing is pretty slick, I like it. Ready to go in as far as I am concerned.

asmeurer:
OK, let's do it.

inducer:
Cool, thanks!

Issue #107, opened 2014-02-21
Ability to sort variables by most recently changed


asmeurer:
It would be nice if the variables view could be sorted with the most recently changed variables at the top, rather than alphabetically. It's often quite hard to follow the view because things change in different places, and most of the time you don't care about most of the variables in the list. 

Something from https://github.com/inducer/pudb/issues/65 may be a prequsite of this.

PR #118, opened 2014-06-10
Allow filter on variables list

albfan:
With non trivial script, variables list becomes a huge list, making it somewhat unmanageable. Sorting by most recent update (another open [issue](https://github.com/inducer/pudb/issues/107)) would do help but It would be cool to allow **"/"** key to start a filter on variables list

Here you are some quick thoughs:
- use \* as regex .*
- use , to search multiple patterns
- use . to search on second level variables (nested)

examples:
- first,last

Show variables **first** and **last**
- current\* 

Show variables like **currentLine**, **currentColumn**, **currentFile**...
- element.pos,.size

Show variable **pos** inside **variable** element, an every **size** variable inside whatever variable

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -94,6 +94,7 @@ Sidebar-related (active in sidebar):
 Keys in variables list:
     \ - expand/collapse
     t/r/s/c - show type/repr/str/custom for this variable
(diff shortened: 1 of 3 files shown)


albfan:
B

PR #120, opened 2014-06-10
Redirect input, output and error at demand

albfan:
I start this feature based on [this conversation](http://lists.tiker.net/pipermail/pudb/2013-May/000174.html). 

I need to debug a python [script which acts like a filter](https://github.com/albfan/filter-word-diff), so I'm interested on provide input and see output while debugging

I don't want to mess my code with std redirections when debugger can be told to do so

to use:

```
$ pudb -i /some/file -o /dev/pts/1 script.py
```

### DIFF
--- a/pudb/__init__.py
+++ b/pudb/__init__.py
@@ -44,8 +44,8 @@ DEFAULT_SIGNAL = signal.SIGINT
 del signal
 
 
-def runscript(mainpyfile, args=None, pre_run="", steal_output=False):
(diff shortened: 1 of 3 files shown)


albfan:
Revisiting this: Take a look at gdb dashboard https://github.com/cyrus-and/gdb-dashboard#display-each-module-in-a-separate-terminal. It is python too, so everything should be reproducible in pudb

Let me know if you still interesting on this

PR #130, opened 2015-01-25
Adding some more emacs bindings to the REPL

FoxLisk:


asmeurer:
The behavior of C-w seems to be wrong if there is trailing space (compared to ipython).

asmeurer:
It would be nice if we could just use readline to get all the bindings in one fell swoop, but this is good enough.

FoxLisk:
Ah, the C-w wasnt really meant to be compatible with anything in particular besides what made sense to me (which is roughly what zsh does). I'll get it more in line with ipython and submit a new patch

FoxLisk:
I believe this emulates ipython's behaviour pretty well now.

inducer:
Thanks for your contribution!

FoxLisk:
My pleasure! Are you going to put a new package up on pypi?

inducer:
Done.

PR #131, opened 2015-02-04
Allow raw mode on rxvt terminals

flupke:
My custom 256 color theme was broken since v2013.5, (more precisely commit afe202cdd09825be5e2ade9cc106845cbabde3a0), because I use rxvt instead of xterm.

This fixes it.


inducer:
Thanks!

PR #132, opened 2015-02-11
Add ctrl-u binding to pudb command line.

alex-netquity:
Adding just a single binding to what was provided in #130: 
* ctrl-u deletes to the beginning of the line in the pudb command line.

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -1506,6 +1506,11 @@ class DebuggerUI(FrameVarInfoKeeper):
             self.cmdline_edit.edit_text = before[i:][::-1] + after
             self.cmdline_edit.edit_post = len(before[i:])
 
+        def cmdline_del_to_start_of_line(w, size, key):
+            pos = self.cmdline_edit.edit_pos
+            self.cmdline_edit.edit_text = self.cmdline_edit.edit_text[pos:]
+            self.cmdline_edit.edit_pos = 0
+
         def toggle_cmdline_focus(w, size, key):
             self.columns.set_focus(self.lhs_col)
             if self.lhs_col.get_focus() is self.cmdline_sigwrap:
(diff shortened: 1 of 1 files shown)


inducer:
Thanks for your contribution!

This doesn't seem to be a conventional emacs binding. (unless I'm wrong) Not that that's bad--it just hurts discoverability a bit if the remaining bindings there are Emacs-ish.

Next, I'm not sure I understand the use case. What exactly are you trying to do that's made easier by this?

alex-netquity:
Ah, my mistake. It's not an emacs binding, it's just a standard bash binding. There's a bit of overlap between the two. It's documented in `man bash readline`:

```
unix-line-discard (C-u)
      Kill backward from point to the beginning of the line.  The killed text is saved on the kill-ring.
```

The use case is simple. The user may type a command that they wish to discard entirely, or discard part of the command they have to the beginning of the line. Part of this is satisfied with the `Clear` button, but that isn't a standard command line interface. It always bugged me that `<C-u>` wasn't available, and when I saw how easy it would be to add it in the above referenced PR, I decided to add it myself.

Note that we don't have a kill-ring in pudb, so that functionality obviously isn't provided by this (very simple) PR.

asmeurer:
C-u is not emacs, but it is standard in readline (including bash and the python interpreter).

alex-netquity:
As @asmeurer outlined in his PR, using readline in the pudb command window would be an ideal solution. I have no idea how to implement that at this point, but I may dig into it should I get some free time. If we did that, people (such as myself) who use vi-mode would also get that in the command window. It would be a more consistent interface.

inducer:
Using readline is not realistic, as Urwid and the curses library behind readline will fight over who's in charge of the screen. But since Ctrl-U seems reasonably standard in a comparable setting, I'm happy to merge your patch.

Issue #103, opened 2014-02-04
Names not defined in list comprehension in the IPython shell


asmeurer:
In the IPython shell, you can't seem to access names from within list comprehensions:

``` py
In [16]: a = 1

In [17]: [a for i in range(10)]
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
/Users/aaronmeurer/Documents/Continuum/conda/conda/resolve.py in <module>()
----> 1 [a for i in range(10)]

/Users/aaronmeurer/Documents/Continuum/conda/conda/resolve.py in <listcomp>(.0)
----> 1 [a for i in range(10)]

NameError: global name 'a' is not defined
```

It doesn't matter if the name was defined within the session or if it comes from the function you are debugging. This is debugging within a function scope, if that matters.

nikhilweee:
4 years later, is there a solution yet?

billtubbs:
Still not solved I guess.  Someone above asked if this occurs in the regular iPython REPL.  It does not.  Here is a related [stack overflow question](https://stackoverflow.com/questions/51164288/python-debugger-sometimes-raises-nameerror-name-is-not-defined) with an example:

in iPython:
```
$ ipython
Python 3.6.5 | packaged by conda-forge | (default, Apr  6 2018, 13:44:09) 
Type 'copyright', 'credits' or 'license' for more information
IPython 6.4.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import pdb; pdb.set_trace()
--Call--
> /Users/billtubbs/anaconda/envs/py36/lib/python3.6/site-packages/IPython/core/displayhook.py(247)__call__()
-> def __call__(self, result=None):
(Pdb) x = 1; [x for i in range(3)]
*** NameError: name 'x' is not defined
```

In regular Python REPL:
```
$ python
Python 3.6.5 | packaged by conda-forge | (default, Apr  6 2018, 13:44:09) 
[GCC 4.2.1 Compatible Apple LLVM 6.1.0 (clang-602.0.53)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import pdb; pdb.set_trace()
--Return--
> <stdin>(1)<module>()->None
(Pdb) x = 1; [x for i in range(3)]
[1, 1, 1]
```

asmeurer:
[...]

PR #133, opened 2015-02-13
Added support for the nifty ptpython shell.

pvaret:
This patch adds support for the awesome [ptpython shell](/jonathanslenders/ptpython).

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -1672,6 +1672,8 @@ class DebuggerUI(FrameVarInfoKeeper):
                 runner = shell.run_ipython_shell
             elif shell.HAVE_BPYTHON and CONFIG["shell"] == "bpython":
                 runner = shell.run_bpython_shell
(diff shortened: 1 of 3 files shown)


asmeurer:
That is a neat shell. I'll give it a try for a while.

Worth noting that this suffers from the same issue as ipython (https://github.com/inducer/pudb/issues/103). If you want to try to fix it, let me know. Otherwise, it looks like this works and it can be merged.

asmeurer:
There's also ptipython (I guess the main advantage is that you get %magics).

inducer:
Thanks for your contribution!

inducer:
Btw, just released this as part of 2015.2.

pvaret:
Re. ptipython: that's a GOOD point. My feeling was that the shell configuration interface is already starting to look a little crowded, and I was not sure the added value of that last option (IPython already being pretty fine on its own) was worth the clutter. It would probably not be too hard to add it as well, however.

As to the variable scoping issue, I'll look into it as time allows. Thanks for pointing it out!

PR #134, opened 2015-02-14
Added toggle sidebar feature

dan-schmierer:
I wanted to add support to copy pudb console output to the clipboard. Unfortunately, python libraries such as tkinter or pyclipboard rely on external modules to provide clipboard support, which makes this feature difficult to provide reliably.

As I work around, I have added support to hide the sidebar using ctrl ^. When the sidebar is not visible you can select multiple lines of text in the console without selecting text from the sidebar.

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -673,6 +673,8 @@ class DebuggerUI(FrameVarInfoKeeper):
             urwid.AttrMap(self.columns, "background"),
             header))
 
+        self._toggle_sidebar(CONFIG["sidebar_visible"])
+
         # }}}
 
         def change_rhs_box(name, index, direction, w, size, key):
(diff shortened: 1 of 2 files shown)


inducer:
Could you add that shortcut to the help screen?

dan-schmierer:
Sure! Is that ok?

inducer:
Now that I've actually tried the patch, I'm actually not completely happy with it. The problem I see is that even in the disabled state, the sidebar can still accept focus, which is confusing. In addition, a small gray bar is still shown on the right.

dan-schmierer:
Good points. I updated the PR to unfocus the sidebar and hide the column separator (dividechars = 0) when the side bar is turned off. What do you think?

inducer:
Not entirely happy--I can still get the focus onto the sidebar, at which point my key events appear to go nowhere... @asmeurer, what do you think?

asmeurer:
I agree we should not allow the keyboard to get to the sidebar if it's hidden. Can you disable key bindings temporarily?

dan-schmierer:
Hi guys, thanks for the feedback!

I know this is not what was suggested, but what about just making the sidebar visible first if someone tries to focus a sidebar element while the sidebar is not visible? Seems like reasonable UI behaviour.

inducer:
You don't get notified when the sidebar gets the focus (IIRC), so you'd have to poll (in the event loop or so). Not sure I'm a fan.

dan-schmierer:
Sidebar focus is set by RHColumnFocuser when one of 'V', 'S', or 'B' is pressed. RHColumnFocuser receives notification of the key press event from SignalWrap.

I just added code inside RHColumnFocuser before it sets focus to the sidebar, to ensure the sidebar is visible first (no polling for focus on the sidebar required). Does this seem reasonable?

inducer:
No, because Urwid also messes with the Focus behind our backs. Pressing "Cursor Right" will also get the sidebar focused.

dan-schmierer:
Aha! Thanks for the clarification, I figured I was missing something. I'll have a think about this.

Issue #137, opened 2015-05-10
I want to see _private members, and not methods


nedbat:
When I examine "self" in the Variables pane, by default it shows me all the methods, which are not useful to me.  And it won't show `_private` names.  I can use "*" to show them, but now I also get `__dunder__` methods which are even more noise and clutter that aren't helping me.

Is there some mode or command I'm missing that lets me see my data without being obscured by methods?

inducer:
Thanks for the suggestion. I'd love to take a PR that addresses this. Here are the relevant bits of code:

* [State](/inducer/pudb/blob/master/pudb/var_view.py#L47): `show_private_members` probably should become a tri-state: No, private, and everything
* [Rendering](/inducer/pudb/blob/master/pudb/var_view.py#L367)
* [UI 1](/inducer/pudb/blob/1b3ee789081f954f99144f3d1bb6f213120b5cb6/pudb/debugger.py#L766) [UI 2](/inducer/pudb/blob/1b3ee789081f954f99144f3d1bb6f213120b5cb6/pudb/debugger.py#L818)

nedbat:
I guess I was imagining the 'm' key would toggle methods or not.  Are you OK with changing the default behavior?  I would prefer that methods not be shown by default.  If so, then '*' could work as you say:  public, public+private, everything.

inducer:
Either of those would work for me.

inducer:
Continuing in #138.

PR #138, opened 2015-05-12
Variables can show/hide methods

nedbat:


nedbat:
This is in response to #137 (is there a way to combine them?)
Not sure I like it yet: it's still annoying that to see `_private`, I have to also see `__doc__`, `__dict__`, etc.  Maybe I will also make '`*`' be a tri-state?

inducer:
I closed #137 in favor of this. I think I would support '*' being tri-state. I don't use private attributes much, but I think the tri-state semantics are what I'd prefer if I did.

nedbat:
OK, I've made '`*`' cycle public/private/all.  Not quite sure how best to describe it in the help, dialog, etc.  Tweaks welcome of course.

inducer:
How about:

```diff
-    * - cycle public/private/dunder members
-    m - toggle methods
+    * - cycle attribute visibility: public/_private/__dunder__
+    m - toggle method visibility
```

There's one slight usability wrinkle with the multi-state visibility, which is that sometimes hitting `*` has no visible effect. I was wondering if a state indicator of some sort might help that.

nedbat:
I see what you mean about `*` not having a visible effect.  Do you have an idea what the indicator would be?

inducer:
I am currently thinking `[*]`/`[_]`/`[__]` in front of the variable name when it's unfolded, but I don't know how much I'd like that in the end. If you've got a suggestion, shoot.

nedbat:
Another option: riffing on the `<omitted private attributes>` that is shown if there is nothing else to show, we could add a line like that when changing visibility wouldn't show anything more.  So if an object has only public data attributes, typing `*` would add a line: `<no private attributes>`, and typing `m` would add `<no methods>`, etc.

inducer:
That marker could be off the screen, though.

nedbat:
Can you show a specific example of what you mean by "`[*] / [_] / [__]` in front of the variable name" ? I'm not sure how you are envisioning it.   `[pub] / [pri] / [all]` might be more understandable, btw.

inducer:
When in the collapsed case some variable might show up as:
```
var_name: SomeClass
```
in the expanded case, I'm imagining
```
var_name: [pub] SomeClass
| attr1: int
```
or
```
var_name: SomeClass [pub]
| attr1: int
```
Btw: yep, `[pub]` et al are better.

nedbat:
OK, I added a marker when values are expanded.  It has two parts, `pub`/`pri`/`all` for access level, and if methods are shown, it adds `+()`, which might be too cryptic.

nedbat:
BTW, I don't understand the urwid stuff enough to know how to make the marker be drawn differently (grayer) than the rest of the label.

inducer:
Thanks very much for your contribution! I'm a huge fan. I've just released 2015.3 with your changes in it.

nedbat:
Yay! :)

Issue #140, opened 2015-06-03
Consider adding support to add count to commands


dohsimpson:
For example, while in a loop, press `50t` to run pass the cursor 49 times before it breaks.
Would be a really nice feature:)

inducer:
Not a bad idea. I unfortunately don't have the bandwidth to do it right now, but I'd certainly review a patch.

PR #144, opened 2015-06-15
added vim-like count feature for certain commands

dohsimpson:
see issue #140
supported commands: run to cursor; movements(j/k/h/l); stack movements(u/d)
Bug: UI is not correctly updated if program terminated after count command

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -1128,6 +1128,53 @@ class DebuggerUI(FrameVarInfoKeeper):
                     self.debugger.set_continue()
                     end()
 
+        def count(w, size, key, pending=0):
+            """vim style add count to key press, currently support:
+            movement keys: k/j/l/h/pg-up/pg-down/u/d
+            run to cursor: t"""
+            if key in list("1234567890"):
+                pending = pending * 10 + int(key)
+                pending_handler = partial(count, pending=pending)
+                # if first time pending
+                if not self.source_sigwrap.pending_handler:
+                    self.source_sigwrap.pending_handler = pending_handler
+                    return None, None
+                return pending_handler, None
+            elif key == "esc":
+                self.source_sigwrap.pending_handler = None
+                return None, None
+            else:
+                try:
+                    handler = [h for (k, h) in self.source_sigwrap.event_listeners if k == key][0]
+                except IndexError:
+                    # self.message("not a valid command: %s" % key)
+                    self.source_sigwrap.pending_handler = None
+                    return None, None
+                else:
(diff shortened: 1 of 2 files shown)


asmeurer:
Is it possible to make it work with arrow keys too?

asmeurer:
Should we add some kind of indicator to the UI so the user can see what number they've typed?

asmeurer:
Seems like `s` and `n` should be supported as well.

pudb/debugger.py
```diff
@@ -1128,6 +1128,53 @@ def run_to_cursor(w, size, key):
...
+                    # if is debugger handler: modify breakpoint ignore count
+                    elif handler in [run_to_cursor]:
+                        result = handler(w, size, key)
+                        # get the breakpoint set by handler
+                        _, pos = self.source.get_focus()
+                        bp_source_identifier = self.source_code_provider.get_breakpoint_source_identifier()
+                        existing_breaks = self.debugger.get_breaks(bp_source_identifier, pos+1)
+                        if existing_breaks:
+                            assert len(existing_breaks) == 1,\
+                                "Expect one breakpoint at line %d, got %d" % (pos+1, len(existing_breaks))
```
asmeurer: I got this error. I think it occurs when you use `t` on a line that is already set as a breakpoint.

dohsimpson: Hi asmeurer, thanks for the feedback. It's not obvious to me how to solve this. The normal breakpoint alway break therefore makes the ignore count of temporary breakpoint useless.
I think that a possible solution would be to disable the normal breakpoint and reenable them after the temporary breakpoints used. But it seems that there's no clean way to do it.

dohsimpson:
The `count` function is only registered in the listener of `DebuggerUI.source_sigwrap`, which doesn't have control of "Up" and "Down". But I think it's doable if we register it as a top level listener.

I agree that UI indicator would be great. But I'm not very familiar with pudb's UI codebase, so I would need some help on that.

Unfortunately `s` and `n` does not use Breakpoint and thus there's no API for adding ignore count.

pudb/debugger.py
asmeurer: I think only temporary breakpoints should be ignored, right? Is there some way to not ignore manually set breakpoints?

dohsimpson: I'm not sure I understand your statement correctly. Manually set breakpoint are never ignored and breaks each time it's met, and therefore would interfere with temporary breakpoint that are set to be ignored for a specified times.

PR #146, opened 2015-06-20
Added ability to save a load saved watch expressions

lechat:
I think that this is useful functionality: key 'z' in 'Variables' view stores all watches to a file, key 'x' loads them.

The save file is stored in the same directory as saved breakpoints.

Saved watches are not autoloaded upon start.

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -100,6 +100,8 @@ Keys in variables list:
     m - toggle method visibility
     w - toggle line wrapping
     n/insert - add new watch expression
+    z - save watch expressions
+    x - load saved watch expressions
     enter - edit options (also to delete)
 
 Keys in stack list:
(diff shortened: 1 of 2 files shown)


inducer:
I like the idea. Why z and x as shortcuts though?

lechat:
Well, obvious "s" for save is already taken... So I chose two unused ones. Don't mind changing them for any other.

inducer:
Not really--"s" is the global thing for 'step', but it Urwid allows each widget to have different shortcuts. Also, another thing: I'd like some amount of user feedback so that they know what their keystrokes did. A message box ("Watch expressions loaded/saved to ...") would do that nicely. Those are easy to create:

```
  self.message("Traceback saved as %s." % fn,
              title="Success")
```

(where `self` is the debugger object)

Another thing that occurred to me is error handling--if something goes wrong with the file I/O on either load or save, the debugger crashes, which I'd like to avoid.

asmeurer:
Why not just save and load them automatically, as is done with breakpoints? This especially doesn't make sense since there is a single global file for saving watch expressions.

lechat:
@inducer "s" is alreade taken in variables view:

```
Keys in variables list:
    \ - expand/collapse
    t/r/s/c - show type/repr/str/custom for this variable
...
```
Good point about error handling, I will add it.

@asmeurer I tried to make it that way, but no luck so far...

inducer:
How come you closed this PR?

lechat:
I decided to rewrite it, so watches will be automatically saved and loaded, same as breakpoints.

PR #148, opened 2015-06-20
Don't call set_continue() in set_trace()

### LINKED ISSUE #147: Unsetting set_trace() continues on recursive call
Consider

```
def test(n):
    import pudb;pudb.set_trace()
    if n > 1:
        return test(n - 1)
    return n

test(3)
```

If you run the file and un-break the `set_trace` line with `b` and press `n`, instead of stepping into the return, the script continues and pudb exists.

asmeurer:
Fixes #147.


inducer:
Looks good.

Issue #145, opened 2015-06-18
pudb barfs if $HOME isn't set


bukzor:
I sometimes need to debug in a reduced-environment ... environment where this variable isn't set.

pudb starts up fine, but subsequently crashes with "Permission denied: /.config".

I'd expect pudb to just skip config persistence in this case, possibly with a warning.

inducer:
Can you cook up a patch?

bukzor:
Almost certainly.

Where do I add the regression test?

inducer:
See #40. :-1:

lechat:
I think that setting default to ``/`` if home is not set on line 4 in settings.py will result in this message.

Possibly falling back to $TEMP first and then to current dir would be the easiest thing to implement.

lechat:
Added pull request for this: #151

bukzor:
Agree. Fallback should be None, and the patch consists primarily of handling the None gracefully.

lechat:
I think that this issue could be closed now.

inducer:
@bukzor : Agree?

PR #151, opened 2015-06-27
Changed finding HOME. Fix for issue #145

lechat:


asmeurer:
I think it would be better to only look for $HOME and if it isn't there, or can't be written to, to completely skip trying to use a config file.

asmeurer:
It's also worth noting that this breaks backwards compabitility.

lechat:
@asmeurer could you elaborate about backwards compatibility? w/o that change pudb just breaks if $HOME is not set.

asmeurer:
The backwards compatibility is broken for the fallback to `/`.

lechat:
I think that it is wrong to fallback to ``/``, no matter what: this is not a place for user apps to write to. This dir rarely have permission for non-elevated user to write.

inducer:
I agree with @asmeurer that a config file should only be looked for in $HOME.

inducer:
Is falling back to the current directory really a good idea, though?

asmeurer:
I don't really see how any of the fallbacks (`/`, `TEMP`, or `.`) are good ideas.

inducer:
Yeah, probably not.

lechat:
Ok. Since that is not a good idea, I am closing this pull requests and submitting another one.

PR #152, opened 2015-07-18
Not saving any settings if HOME and XDG_CONFIG_HOME are not set

lechat:
Fix for issue #145 

pudb will not break if HOME is not set and still try to use XDG_CONFIG_HOME. If both are not set - no settings will be saved.

I also added description about environment variables used to calculate path to store configuration/breakpoints.


inducer:
Looks good to me. Thanks for your contribution!

PR #153, opened 2015-07-24
Expand variables for custom theme path

amigrave:
For those who want to use environment variables (eg: $XDG_CONFIG_HOME) in a custom theme's path.


inducer:
Good idea, thanks!

PR #154, opened 2015-07-31
add agr-256 theme

amigrave:
This 256 colors dark theme is easy on the eyes and allows you to debug all night long !

Screenshot


inducer:
Thanks!

PR #159, opened 2015-12-04
fix module loading on pudb2

albfan:


setup.py
```diff
@@ -49,6 +49,6 @@
           ],
       packages=["pudb"],
       entry_points={
-          'console_scripts': ['pudb' + PY_VERSION + ' = pudb.run:main'],
+          'console_scripts': ['pudb2' + PY_VERSION + ' = pudb.run:main'],
```
asmeurer: This is wrong. The PY_VERSION is already added to "pudb".

setup.py
```diff
@@ -16,7 +16,7 @@
 finally:
     readme.close()
 
-setup(name='pudb',
+setup(name='pudb2',
```
asmeurer: The package name should be pudb, not pudb2.

asmeurer:
Can you specify what bug this fixes, and how to reproduce it?

setup.py
albfan: Sorry, just want to PR second commit. My fault

albfan:
Sorry this was my very own tag. Have to check tip

PR #163, opened 2015-12-31
regularlize whitespace for multiline strings

jsoref:
Include a trailing space on the first line.
Omit leading space on following line.


inducer:
Thx.

PR #174, opened 2016-03-25
Ensure var view never splits Unicode chars

pquentin:
TextCanvas wants byte counts, but we were sending column counts. It does
fix the immediate issue, but when a given variable is so long that we
need to scroll, there are now some dangling characters after the final
quote, even when the variable only contains ASCII.

This imports a private variable from urwid.util to avoid depending too
much on apply_target_encoding magic - not sure if a good idea.

### DIFF
--- a/pudb/ui_tools.py
+++ b/pudb/ui_tools.py
@@ -1,4 +1,5 @@
 import urwid
+from urwid.util import _target_encoding
 
 
 # generic urwid helpers -------------------------------------------------------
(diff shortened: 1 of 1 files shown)


inducer:
@asmeurer You wrote the original wrapping code. Could you take a look?

asmeurer:
Anyway, it seems to fix the issue.

asmeurer:
I don't know what the risks are for using a private variable from urwid. Other than that, it looks like a good fix.

inducer:
Agree, looks good. Thanks, both of you!

pquentin:
Thanks!

PR #175, opened 2016-03-31
Fix make_canvas for lines with multiple attrs

pquentin:
As I was saying in #174, I was seeing some dangling characters issue. The issue was more severe than that, because I was always using `line[i:count]` instead of `line[i:i+count]`. To ensure this two chars fix works I added unit tests. They can be run by using 'py.test' from the root pudb directory.

### DIFF
--- a/pudb/ui_tools.py
+++ b/pudb/ui_tools.py
@@ -30,7 +30,7 @@ def make_canvas(txt, attr, maxcol, fill_attr=None):
         def get_byte_line_attr(line, line_attr):
             i = 0
             for label, column_count in line_attr:
-                byte_count = len(line[i:column_count].encode(_target_encoding))
+                byte_count = len(line[i:i+column_count].encode(_target_encoding))
                 i += column_count
                 yield label, byte_count
 
--- a/unittests/test_make_canvas.py
+++ b/unittests/test_make_canvas.py
@@ -1,0 +1,45 @@
+from pudb.ui_tools import make_canvas
+
+class TestMakeCanvas():
+    def test_simple(self):
+        text = u'aaaaaa'
+        canvas = make_canvas(
+            txt=[text],
+            attr=[[('var value', len(text))]],
+            maxcol=len(text) + 5
+        )
+        content = list(canvas.content())
+        assert content == [
+            [('var value', None, b'aaaaaa'), (None, None, b' ' * 5)]
+        ]
+
+    def test_multiple(self):
+        canvas = make_canvas(
+            txt=[u'Return: None'],
+            attr=[[('return label', 8), ('return value', 4)]],
+            maxcol=100
+        )
+        content = list(canvas.content())
+        assert content == [
+            [('return label', None, b'Return: '),
+             ('return value', None, b'None'),
+             (None, None, b' ' * 88)]
+        ]
+
+    def test_boundary(self):
+        text = u'aaaaaa'
+        canvas = make_canvas(
+            txt=[text],
+            attr=[[('var value', len(text))]],
+            maxcol=len(text)
+        )
+        assert list(canvas.content()) == [[('var value', None, b'aaaaaa')]]
+
+    def test_byte_boundary(self):
+        text = u'aaaaaaé'
+        canvas = make_canvas(
+            txt=[text],
+            attr=[[('var value', len(text))]],
+            maxcol=len(text)
+        )
+        assert list(canvas.content()) == [[('var value', None, b'aaaaaa\xc3\xa9')]]


inducer:
Thanks!

inducer:
These tests fail for me:

```
  File "/home/andreas/work/pudb/test/test_make_canvas.py", line 50, in test_byte_boundary
    assert list(canvas.content()) == [[('var value', None, b'aaaaaa\xc3\xa9')]]
AssertionError: assert [[('var value...one, '\xa9')]] == [[('var value'...aa\xc3\xa9')]]
  At index 0 diff: [('var value', None, 'aaaaaa\xc3'), (None, None, '\xa9')] != [('var value', None, 'aaaaaa\xc3\xa9')]
  Use -v to get the full diff
```

pquentin:
Can you share `python -c 'import locale; print(locale.setlocale(locale.LC_ALL, ""))'`? My test expects UTF-8.

pquentin:
Two py.test notes:

 * you don't need to add `__main__` to test, selecting a file/class/test can be done with `-k make_canvas`. 
 * consider using `test_requires` so that pudb users won't be forced to install py.test.

inducer:
> you don't need to add __main__ to test, selecting a file/class/test can be done with -k make_canvas. 

Oh, TIL. Still, sometimes I like to pass arguments to tests, which that doesn't seem to do easily. Doesn't hurt much, either.

> consider using test_requires so that pudb users won't be forced to install py.test.

Thx, done.

inducer:
`python -c 'import locale; print(locale.setlocale(locale.LC_ALL, ""))'` gives:
```
de_DE.utf8
```

pquentin:
Not sure what's going on here. Could you please run the following code (taken directly from urwid/util.py)?

```
def detect_encoding():
    # Try to determine if using a supported double-byte encoding
    import locale
    try:
        try:
            locale.setlocale(locale.LC_ALL, "")
        except locale.Error:
            pass
        return locale.getlocale()[1] or ""
    except ValueError as e:
        # with invalid LANG value python will throw ValueError
        if e.args and e.args[0].startswith("unknown locale"):
            return ""
        else:
            raise

print(detect_encoding())
```

It prints "UTF-8" for me.

inducer:
Prints `UTF-8` for me.

pquentin:
This is really surprising and difficult to debug via GitHub :-)

- Can you share the whole test session (to show Python/pytest versions)
- Can you try with Python 3?
- Can you try with https://github.com/pquentin/pudb/commit/17c12bb42b98b5335e828bbb2c19ac8d0dae20f5 (even if I don't think it will change anything)?

Thank you and sorry for the trouble.

pquentin:
Trying with Python 3 could be helpful because we would rule out a few issues this way. (It does work with both Python 2 and 3 here.)

inducer:
Py3 seems to work. Py2 fails with:

```
========================================================= test session starts ==========================================================
platform linux2 -- Python 2.7.8, pytest-2.9.1, py-1.4.31, pluggy-0.3.1
rootdir: /home/andreas, inifile: pytest.ini
collected 4 items 

test_make_canvas.py ...F

=============================================================== FAILURES ===============================================================
__________________________________________________________ test_byte_boundary __________________________________________________________
  File "/home/andreas/work/pudb/test/test_make_canvas.py", line 50, in test_byte_boundary
    assert list(canvas.content()) == [[('var value', None, b'aaaaaa\xc3\xa9')]]
AssertionError: assert [[('var value...one, '\xa9')]] == [[('var value'...aa\xc3\xa9')]]
  At index 0 diff: [('var value', None, 'aaaaaa\xc3'), (None, None, '\xa9')] != [('var value', None, 'aaaaaa\xc3\xa9')]
  Use -v to get the full diff
================================================== 1 failed, 3 passed in 0.01 seconds ==================================================
```

https://github.com/pquentin/pudb/commit/17c12bb42b98b5335e828bbb2c19ac8d0dae20f5 appears to die with

```
========================================================= test session starts ==========================================================
platform linux2 -- Python 2.7.8, pytest-2.9.1, py-1.4.31, pluggy-0.3.1
rootdir: /home/andreas, inifile: pytest.ini
collected 0 items / 1 errors 

================================================================ ERRORS ================================================================
_________________________________________ ERROR collecting work/pudb/test/test_make_canvas.py __________________________________________
test_make_canvas.py:3: in <module>
    from pudb.ui_tools import make_canvas, _target_encoding
E   ImportError: cannot import name _target_encoding
======================================================= 1 error in 0.08 seconds ========================================================
```
on Py2 (fine with Py3).

pquentin:
@inducer This appears to suggest that your Python 2 environment does not take into account the contents from the working tree. Maybe run `pip2 uninstall pudb` multiple times until there's nothing left and then run `python2 setup.py develop` to ensure you're really using the latest commit?

inducer:
Uh-oh. I have to apologize. That was totally my fauilt. You're right. `master` passes in Py2. Sorry for causing you grief, and thanks for your patience.

pquentin:
No worries, I'm glad we sorted this out.

Issue #177, opened 2016-04-07
pudb isn't compatible w/ mercurial.demandimport


jsoref:
Mercurial has a magical import wrapper called demandimport.

```py
import mercurial
import mercurial.demandimport
mercurial.demandimport.enable()
import foo
```

if foo does not exist, it will not throw an exception at this point.
Instead, it will throw its exception when you do:

```py
foo.bar
```

I realized this when using **py3** (not **IronPython**) w/ mercurial.demandimport and **pudb3**:
```py
>>> ctypes.util                                                                                    
Traceback (most recent call last):                                                                 
  File "/home/timeless/hg/py3/lib/python3.5/site-packages/pudb/debugger.py", line 1415, in         
cmdline_get_namespace                                                                              
    from pudb.shell import SetPropagatingDict                                                      
  File "/home/timeless/hg/crew/mercurial/demandimport.py", line 143, in _demandimport              
    return _hgextimport(_import, name, globals, locals, fromlist, level)                           
  File "/home/timeless/hg/crew/mercurial/demandimport.py", line 53, in _hgextimport                
    return importfunc(name, globals, *args, **kwargs)                                              
  File "/home/timeless/hg/py3/lib/python3.5/site-packages/pudb/shell.py", line 183, in <module>    
    if HAVE_IPYTHON and hasattr(IPython, 'Shell'):                                                 
  File "/home/timeless/hg/crew/mercurial/demandimport.py", line 132, in __getattribute__           
    self._load()                                                                                   
  File "/home/timeless/hg/crew/mercurial/demandimport.py", line 98, in _load                       
    mod = _hgextimport(_import, head, globals, locals, None, level)                                
  File "/home/timeless/hg/crew/mercurial/demandimport.py", line 53, in _hgextimport                
    return importfunc(name, globals, *args, **kwargs)                                              
ImportError: No module named 'IPython'                                                             
```

The fix for this sort of thing is to access a property of each module inside the try block that intends to import the module. I'll try to write a PR for this, but first I'll need to find out what properties IPython has (and probably try installing it...

asmeurer:
[...]

PR #179, opened 2016-04-14
access module properties to ensure modules are present

jsoref:
demandimport wrappers such as mercurial prevent simple imports from triggering exceptions.
By accessing a property, the modules are actually loaded (or fail).

This should fix issue #177

### DIFF
--- a/pudb/shell.py
+++ b/pudb/shell.py
@@ -1,5 +1,9 @@
 try:
     import IPython
+    # Access a property to verify module exists in case
+    # there's a demand loader wrapping module imports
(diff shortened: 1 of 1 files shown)


inducer:
I agree with @asmeurer's assessment in #177--this isn't really our bug. But then, this workaround doesn't hurt anything on our side, so there's no reason to not accept it. One small request: Could you add some comment lines to explain why those attribute accesses are there? Github URL is fine.

jsoref:
Updated

inducer:
Thanks!

Issue #178, opened 2016-04-14
get_executable_lines_for_file crashes for not compiling files


jsoref:
I'm in the middle of merging mercurial, so I have merge conflicts. I also am debugging it (albeit with a different hg).

```
(py)[timeless@gcc2-power8 crew8]$ ./hg --cwd ../crew summary --debugger
entering debugger - type c to continue starting hg or h for help
Traceback (most recent call last):
  File "/home/timeless/hg/crew8/mercurial/dispatch.py", line 202, in _runcatch
    debugtrace[debugger]()
  File "/home/timeless/hg/py/lib/python2.7/site-packages/pudb-2015.4.2-py2.7.egg/pudb/__init__.py", line 150, in set_trace
    dbg = _get_debugger()
  File "/home/timeless/hg/py/lib/python2.7/site-packages/pudb-2015.4.2-py2.7.egg/pudb/__init__.py", line 35, in _get_debugger
    dbg = Debugger(**kwargs)
  File "/home/timeless/hg/py/lib/python2.7/site-packages/pudb-2015.4.2-py2.7.egg/pudb/debugger.py", line 165, in __init__
    for bpoint_descr in load_breakpoints():
  File "/home/timeless/hg/py/lib/python2.7/site-packages/pudb-2015.4.2-py2.7.egg/pudb/settings.py", line 482, in load_breakpoints
    return parse_breakpoints(lines)
  File "/home/timeless/hg/py/lib/python2.7/site-packages/pudb-2015.4.2-py2.7.egg/pudb/settings.py", line 448, in parse_breakpoints
    if get_breakpoint_invalid_reason(filename, lineno) is None:
  File "/home/timeless/hg/py/lib/python2.7/site-packages/pudb-2015.4.2-py2.7.egg/pudb/lowlevel.py", line 46, in get_breakpoint_invalid_reason
    if lineno not in get_executable_lines_for_file(filename):
  File "/home/timeless/hg/py/lib/python2.7/site-packages/pudb-2015.4.2-py2.7.egg/pudb/lowlevel.py", line 23, in get_executable_lines_for_file
    codes = [compile("".join(getlines(filename)), filename, "exec")]
  File "/home/timeless/hg/crew/mercurial/util.py", line 443
    <<<<<<< local
     ^
SyntaxError: invalid syntax
** unknown exception encountered, please report by visiting
** https://mercurial-scm.org/wiki/BugTracker
[...]

PR #180, opened 2016-04-14
get_executable_lines_for_file: tolerate files that do not compile

jsoref:
Fixes issue #178


inducer:
Minor wrinkle: You probably want to use the result in `execable_lines` in the `if` statement just below. Looks good otherwise.

asmeurer:
Yep, agree with that.

jsoref:
Oops, yes, I meant to do that.

jsoref:
(Done)

inducer:
Thanks for your contribution!

PR #184, opened 2016-05-24
allow to set the host and port in pudb.remote.set_trace()

arne-cl:
This patch allows to use set_trace() when pudb runs on another machine (e.g. a VM).


inducer:
Looks good. Thanks for the contribution!

PR #185, opened 2016-05-30
Display an X next to disabled breakpoints

asmeurer:
I also plan on adding a theme color for this.  Let me know how you feel about the `X` and the `# hits` UI (I'm happy to change/remove either).

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -1210,8 +1210,16 @@ class DebuggerUI(FrameVarInfoKeeper):
                 existing_breaks = self.debugger.get_breaks(
                         bp_source_identifier, lineno)
                 if existing_breaks:
-                    err = self.debugger.clear_break(bp_source_identifier, lineno)
-                    sline.set_breakpoint(False)
+                    err = None
+                    for bp in existing_breaks:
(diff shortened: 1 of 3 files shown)


asmeurer:
This is ready to be reviewed.

inducer:
Looks good, feel free to merge.

pudb/debugger.py
```diff
@@ -546,6 +546,8 @@ def get_lines(self, debugger_ui):
             return [SourceLine(debugger_ui, self.file_name)]
 
         breakpoints = debugger_ui.debugger.get_file_breaks(self.file_name)[:]
+        breakpoints = [lineno for lineno in breakpoints if
```
asmeurer: I think this line is unnecessary. It might be necessary if we ever save the "enabled" state to the breakpoints file, however.

asmeurer:
Would love to hear folks' opinions on my choice of theme colors for disabled breakpoints. I am happy with the "midnight" colors, but that's the only theme I use. The rest I tried to find something that looks OK, but if you use the theme and want them to look better, let me know.

inducer:
I'll change them once they bother me. :)

Issue #194, opened 2016-08-03
Enable `from __future__ import absolute_import` to work around bug on non case-sensitive file systems


rory-geoghegan-ecometrica:
I hit a bug where in python 2.7, the `pudb/shell.py` module will try to import (innocently) the IPython package (to set the `HAVE_IPYTHON` variable), but actually import the `pudb/ipython.py` module instead, because my filesystem is case-insensitive inside my docker/docker-machine setup and it is treating `import IPython` as a relative import to the `pudb` package.

The most visible issue is my case-insensitive file-system, but also python 2.7 is trying to do a relative import in this case, while it really should not, as it does not by default in python 3.

Can we add the `from __future__ import absolute_import` either to just that module or to every module? According to [the pep](https://docs.python.org/2.5/whatsnew/pep-328.html), it should be supported by the python 2.6+ versions you support in pudb.

What would be easiest for you? Should I create a pull request?

inducer:
PR sounds good.

asmeurer:
I personally prefer to put `from __future__ import absolute_import, division, print_function` at the top of every file to avoid any possible issues like this.

inducer:
Also fine by me, now that 2.5 is no longer an issue.

rory-geoghegan-ecometrica:
Done: https://github.com/inducer/pudb/pull/195

PR #195, opened 2016-08-04
Force absolute imports, division and print function across all modules.

rory-geoghegan-ecometrica:
With the absolute imports, this fixes an issue with pudb/shell.py where it
was confusing importing the IPython package with the pudb.ipython module,
as detailed in the [github case
here](https://github.com/inducer/pudb/issues/194).


inducer:
Thanks for your contribution!

PR #199, opened 2016-08-30
Fix module reloading

asmeurer:
The Python 2/3 check was reversed from what it was supposed to be.

FWIW, the reload button is confusing. It reloads the selected module, even though it isn't selected any more if you arrow over to button. I would have expected it to reload the list.  And what is the benefit of reloading a module anyway?

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -1375,10 +1375,10 @@ class DebuggerUI(FrameVarInfoKeeper):
                         mod_name = widget.base_widget.get_text()[0]
                         mod = sys.modules[mod_name]
                         if PY3:
-                            reload(mod)  # noqa (undef on Py3)
-                        else:
                             import importlib
                             importlib.reload(mod)
+                        else:
+                            reload(mod)  # noqa (undef on Py3)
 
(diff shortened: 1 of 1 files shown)


inducer:
Thanks for the patch!

> And what is the benefit of reloading a module anyway?

That was based on the notion that you may want to fix a bug and reload the changed code. In six years of using pudb (has it been that long?!) I've never used that once. So I wouldn't be opposed to getting rid of it.

asmeurer:
I've never used it either. Reloading modules can be dangerous, as it can break a lot of things. For instance, if a module defines a class, reloading it will define the class again, meaning any existing instances will no longer be instances of the new class (see also https://docs.python.org/3/library/importlib.html#importlib.reload). 

Obviously, dangerous things can be OK when debugging sometimes, but I'm not sure about this one. I personally would be too worried about this introducing weird "fake" bugs while debugging to ever used it. But I also don't generally debug things that are a pain to restart completely, so maybe others could find it useful. With that being said, you could always do this manually if really wanted to from the Python console.

asmeurer:
And either way, it would be nice to fix the UI here, if possible.

PR #200, opened 2016-08-31
redraw the whole screen when switching themes

mdamien:
Currently, when selecting a new theme in the settings menu, only two lines get updated with the new theme. This redraw the whole screen with the new theme.


inducer:
Thanks for your contribution!

PR #204, opened 2016-10-12
Updating access to bpython version

thornycrackers:
I don't know if this the correct way to do this but to get bpython working with my version of pudb I had to change this variable or else I was getting errors saying:

```
bpython.version
AttributeError: 'module' object has no attribute 'version'
```

After changing the snippet everything seemed to work fine.

### DIFF
--- a/pudb/.vimcache/ctrlp/mru/cache.txt
+++ b/pudb/.vimcache/ctrlp/mru/cache.txt
@@ -1,0 +1,1 @@
+/work/shell.py
--- a/pudb/.vimcache/ctrlp/mru/cache.txt
+++ b/pudb/.vimcache/ctrlp/mru/cache.txt
@@ -1,1 +1,0 @@
-/work/shell.py
--- a/pudb/shell.py
+++ b/pudb/shell.py
@@ -5,7 +5,10 @@ try:
     # Access a property to verify module exists in case
     # there's a demand loader wrapping module imports
     # See https://github.com/inducer/pudb/issues/177
-    bpython.__version__
+    if hasattr(bpython, 'version'):
+        bpython.version
+    else:
+        bpython.__version__
 except ImportError:
     HAVE_BPYTHON = False
 else:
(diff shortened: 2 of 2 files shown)


inducer:
Thanks for your patch. The purpose of the code you modified is to see if bpython is available. Could you modify it so that this succeeds for both the old behavior (defining `.version`) and the new (defining (`.__version__`)?

thornycrackers:
Does just importing bpython suffice to check if it's available? The line with the version doesn't get assigned or used so really couldn't we just delete it all together?

inducer:
No. See #177.

thornycrackers:
Derp, it's right there in the code. I just used hasattr to check which I think should be good enough.

inducer:
Not sure what problem you're having, but according to:

https://github.com/bpython/bpython/blob/master/bpython/__init__.py

both `version` and `__version__` should be available. This is looking like it's something on your end. (And FWIW, I don't trust these idiotic demand loader things to even get `hasattr` right.)

thornycrackers:
I pip installed pudb and then tried to access the command line (Ctrl - x) and every time I try to it gives me an error on this line saying the attribute doesn't exist. I know both should be available but it doesn't work and I have to change these lines. If you don't like hasattr catch a AttributeError exception and then set HAS_BPYTHON to true

inducer:
What happens if you run

```
import bpython
bpython.version
```

and 

```
import bpython
bpython.__version__
```

outside of pudb?

thornycrackers:
2__tmux

Here is a screenshot

inducer:
Fair enough:

https://github.com/bpython/bpython/blob/742a2d2af0bdcc7b65890eb918ecfca28d7da01a/bpython/__init__.py

Is there  a reason why you're staying on a release of bpython that's  >4 years old?

asmeurer:
It seems like `bpython.version` is an accidental API https://github.com/bpython/bpython/commit/74ce9642236dd4be23a533ec123cef7d71eaca3b. `bpython.__version__` should likely be preferred.

inducer:
Ok, you've convinced me. I've effectively merged the original version of @thornycrackers's patch. @thornycrackers, sorry for giving you the runaround on this.

Issue #207, opened 2016-12-09
Monokai Theme


epmoyer:
I created a Monokai theme for PuDB (https://github.com/epmoyer/pudb-theme-monokai). Would you be interested in a pull request that adds Monokai to the built-in-themes? Or do you have a different vision for aggregating custom themes over time (like putting them in a themes directory?)

Cannonical Monokai colorizes some source elements which are not exposed by the current palette API (like the "def" statement, and function arguments).  Would you be interested in a pull request that exposes a couple new source elements such that Monokai could be styled more faithfully?

I've fallen in love with PuDB; thanks for all your hard work! And thanks for supporting custom themes.

jeffwidman:
Personally, as a user I'd like to see both of these--I'm a fan of Monokai, plus colorizing the other source elements.

asmeurer:
The current theme list is short, and Monokai is popular. I would go ahead and add it. Also fine with more complete lexing.

inducer:
Both sound good. Submit away.

inducer:
(And thanks for offering to contribute!)

epmoyer:
Status report:

I've been successful at extracting the necessary tokens from Pygments to implement a faithful representation of Monokai (with the exception of function arguments).  It's not ready for prime-time yet, but I thought I'd post a quick screenshot comparison of the current Monokai rendering in pudb vs. Sublime.  I decided to implement both an ANSII color and Xterm color version (the ANSII version is shown).

The remaining hurdle is to see if I can't detect function arguments somehow (Monokai colors them orange).  Pygments does not tokenize function arguments in any special way, so a small bit of context sensitive parsing will be required to tease them out.  Feeling bullish.

I don't want to push the branch yet because the current changes have side effects on the legacy styles and I haven't fixed that yet. 

<img width="1180" alt="monokai_progress" src="https://cloud.githubusercontent.com/assets/136718/21082472/1c632a40-bf91-11e6-85a8-a96cb7afc9de.png">

Cheers!

epmoyer:
I am closing this issue.  Pull request submitted.

jeffwidman:
Just a thought: 
The more typical workflow I've seen in open source projects on github is to leave the issue open and in the PR add the words "Fix #207". Github will auto-crosslink the issue/PR and when/if the PR is accepted Github will automatically mark the issue as closed. It help decouple the high-level product design/engineering architecture discussion in the issue from the line-by-line engineer implementation discussion of the PR.

[...]

PR #208, opened 2016-12-12
Monokai

epmoyer:
This change adds "monokai" and "monokai-256" to the built-in Themes. It faithfully reproduces all Monokai styling except for coloring optional argument names orange in function calls (implementing that would have required a two-pass parser to go backwards and change the token type of the argument name after encountering the "=" which follows it.)

This implementation adds several new palette items: "namespace", "operator", "argument", "builtin", "pseudo", "dunder", "exception", and "keyword2".  Their meaning is documented near the top of theme.py

All legacy Themes have been updated to assign colors to the new palette items such that the legacy themes appear identical before and after this change.  The addition of the new palette items makes it possible, if you like, to expand the colorization of the legacy themes in the future. I thought it was cleanest for me to keep this pull request focussed on the new theme and not muddy the waters by making any changes to the appearance of the legacy ones.

I have tested on Python 3.5 and 2.7, and raw/curses (though both appear identical on my platform).

**monokai-256 vs Sublime**
monokai-256

**monokai vs Sublime**
monokai

**Legacy themes with Monokai build (left) and trunk build (right)**
classic_compare

vim_compare

dark_vim_compare

midnight_compare

solarized_compare

agr-256_compare

### DIFF
--- a/pudb/source_view.py
+++ b/pudb/source_view.py
@@ -157,9 +157,11 @@ def format_source(debugger_ui, lines, breakpoints):
                 t.Token: "source",
                 t.Keyword.Namespace: "namespace",
                 t.Token.Argument: "argument",
+                t.Token.Dunder: "dunder",
+                t.Token.Keyword2: 'keyword2',
                 t.Keyword: "keyword",
                 t.Literal: "literal",
-                t.Name.Exception: "keyword",
+                t.Name.Exception: "exception",
                 t.Name.Function: "name",
                 t.Name.Class: "name",
                 t.Name.Builtin: "builtin",
@@ -178,13 +180,11 @@ def format_source(debugger_ui, lines, breakpoints):
 
         ATTR_TRANSLATE = {
                 t.Keyword: {
-                    'try': t.Operator.Word,
-                    'except': t.Operator.Word,
-                    'pass': t.Operator.Word,
-                    'return': t.Operator.Word,
-                    'for': t.Operator.Word,
-                    'if': t.Operator.Word,
-                    'yield': t.Operator.Word,
+                    'class': t.Token.Keyword2,
+                    'def': t.Token.Keyword2,
+                    'exec': t.Token.Keyword2,
+                    'lambda': t.Token.Keyword2,
+                    'print': t.Token.Keyword2,
                     },
                 t.Operator:{
                     '.': t.Token,
@@ -195,9 +195,6 @@ def format_source(debugger_ui, lines, breakpoints):
                 t.Name.Builtin:{
                     'object': t.Name.Class,
                     },
-                # t.Name:{
-                #     'a': t.Token.Argument, # For Debug testing of argument parsing
-                #     }
                 }
 
         class UrwidFormatter(Formatter):
@@ -214,7 +211,7 @@ def format_source(debugger_ui, lines, breakpoints):
 
                     # Find function arguments. When found, set their
                     # ttype to t.Token.Argument
-                    new_ttype= argument_parser.parse_token(ttype, s)
+                    new_ttype = argument_parser.parse_token(ttype, s)
                     if new_ttype:
                         ttype = new_ttype
 
@@ -224,7 +221,7 @@ def format_source(debugger_ui, lines, breakpoints):
                             ttype = ATTR_TRANSLATE[ttype][s]
                     # Translate dunder methods to buitins
                     if ttype == t.Name.Function and s.startswith('__') and s.endswith('__'):
-                        ttype = t.Name.Builtin
+                        ttype = t.Token.Dunder
 
                     while not ttype in ATTR_MAP:
                         if ttype.parent is not None:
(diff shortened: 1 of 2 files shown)


epmoyer:
Ack! Missed a dunder style in solarized and just noticed it in my screenshots.  Fixed now.

solarized_compare_2

jeffwidman:
Fix #207

pudb/source_view.py
```diff
@@ -222,3 +278,51 @@ def shipout_line():
                 PythonLexer(stripnl=False), UrwidFormatter())
 
         return result
+
+from enum import Enum
+class ParseState(Enum):
```
asmeurer: Enum is Python 3-only, but pudb needs to support Python 2 for the moment. I think this should still work just fine if you make it a regular class for how you are using it.

pudb/source_view.py
```diff
@@ -222,3 +278,51 @@ def shipout_line():
...
+    '''States for the ArgumentParser class'''
+    idle = 1
+    found_function = 2
+    found_open_paren = 3
+
+class AgumentParser():
```
asmeurer: Subclass from `object` for Python 2 compatibility.

asmeurer:
Do the new tokens fallback to their counterparts automatically? If so, then the changes to the existing themes seems redundant. If not, then what will happen if someone has their own custom theme?

asmeurer:
I'm unclear if it's possible with the current code, but terminals can support italics (see https://web.archive.org/web/20161003094535/https://alexpearce.me/2014/05/italics-in-iterm2-vim-tmux/ for iTerm2, for instance).

pudb/source_view.py
epmoyer: Agreed!  I now have a 2.6 environment setup and have tested removing the enum with 2.6 & 3.5 (works).  Will push in a moment.

pudb/source_view.py
epmoyer: Agreed

epmoyer:
The new palette items do not automatically inherit from the old set, but I definitely see your point about the desirability of supporting existing custom themes without modification.  It seems achievable, since there is a direct inheritance mapping which would preserve the look of existing themes.  I'll take a crack at it tomorrow.  It's a fine idea, thanks!

And you're right, if the inheritance mapping was automatic then I could remove the new palette items from the legacy themes.

epmoyer:
The current release of urwid (1.3.1) does not contain support for italics.  Support for italics was committed on Apr 19, 2015 (and is part of the master branch now), but has not yet made it into a release (1.3.1 was released Nov 1, 2015, but did not include the italics commit).  I found that Apple’s terminal app does not support italics, but iTerm2 does (I was able to see italics there by printing the associated escape codes), so that’s a viable test environment whenever the next urwid release drops.

This prints italics in terminals that support it:
`print ('plain text \033[3m italic text \033[0m plain text’)`

And this will print in italics when the next version of urwid comes out:
```
import urwid

txt = urwid.Text(("mystyle", "Hello World"))
fill = urwid.Filler(txt, 'top')
palette = [('mystyle', 'default,italics', 'default'),]
loop = urwid.MainLoop(fill, palette)
loop.run()
```

asmeurer:
No need to necessarily add italics. Just wanted to point out that it was possible if you wanted to do it. I guess we should wait for urwid support.

epmoyer:
Added an inheritance mapping which assigns each new added style to the value of it's corresponding legacy style if a Theme does not assign that style explicitly.  With that in place, I regressed all the legacy themes to their original code (and they work properly).  The inheritance is applied after custom Themes are loaded, so custom Themes produce identical visual behavior without modification.

I also updated "monokai" to use a dark style for the side panels (like "monokai-256" does).

**monokai**
monokai

**monokai-256**
monokai-256

**monokai build (left) vs master build (right) with an (ugly :) unmodified legacy custom theme**
compare_custom

**monokai build (left) vs master build (right) with legacy themes**
compare_classic

compare_vim

compare_dark_vim

comare_midnight

compare_solarized

compare_agr-256

pudb/source_view.py
```diff
@@ -151,14 +151,34 @@ def format_source(debugger_ui, lines, breakpoints):
         import pygments.token as t
 
         result = []
-
+        argument_parser = AgumentParser(t)
```
inducer: `Agument` -> `Argument`

pudb/source_view.py
```diff
@@ -222,3 +278,50 @@ def shipout_line():
...
+    idle = 1
+    found_function = 2
+    found_open_paren = 3
+
+class AgumentParser(object):
+    '''Parse source code tokens and identify function arguments.
+
+    This parser implements a state machine which accepts
+    Pygments tokens, delivered sequentially from the beginning
+    of a source file to it's end.
```
inducer: it's -> its

inducer:
Looks great. Thanks for contributing this!

epmoyer:
My pleasure! Thanks Andreas for making such an awesome tool and for being so responsive.  And thanks Aaron for the excellent comments & responsiveness; the patch ended up much better for your suggestions.

Cheers everyone!

jeffwidman:
Thanks @epmoyer really high quality PR that I'm looking forward to using

PR #209, opened 2016-12-15
Setting term_size to the current size if None

timgabets:
This improvement may be useful for the 'remote' debugging in a next terminal tab - in this case the debugger window will be opened with a full screen resolution, not with 80x24

### DIFF
--- a/pudb/remote.py
+++ b/pudb/remote.py
@@ -9,8 +9,10 @@ import errno
 import os
 import socket
 import sys
+import shutil
 
 from pudb.debugger import Debugger
+from pudb.py3compat import PY3
 
 __all__ = ['PUDB_RDB_HOST', 'PUDB_RDB_PORT', 'default_port',
            'debugger', 'set_trace']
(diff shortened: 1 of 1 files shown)


inducer:
Thanks for your contribution!

I'd like to see more error handling here, so that even if this fails, it doesn't make matters worse. I'm also not a fan of doing this via a subprocess, since there are many situation where forking a aubprocess is undefined behavior, such as when a GPU context is active, or when the process is part of an MPI communicator that's backed by a DMA-based network such as Infiniband. Coincidentally, these are primary uses (for me) of remote debugging. I'd prefer this [IOCTL-based solution](http://stackoverflow.com/a/566752).

Lastly, if available, this should just use [Python's built-in facility](https://docs.python.org/3/library/shutil.html#shutil.get_terminal_size) to get the terminal size.

timgabets:
Sorry for a late reply. 

> Lastly, if available, this should just use Python's built-in facility to get the terminal size.

Umm, well, as I see, shutil.get_terminal_size() is only available for Python 3. Checking a python version inside of the set_trace() looks a little ugly to me, but if it's OK to you, I'll do it.

inducer:
> Umm, well, as I see, shutil.get_terminal_size() is only available for
> Python 3. Checking a python version inside of the set_trace() looks a
> little ugly to me, but if it's OK to you, I'll do it.

That's just the cost of doing business in Py2 and 3, to my mind at least.

timgabets:
Done

inducer:
Thanks!

PR #210, opened 2016-12-24
Drop support for ancient iPython versions

jeffwidman:
We are now on iPython 5, version 1 is over 3 years old, so no sense supporting old versions at this point. If someone needs to support something that old, they can use an old version of `pudb`.


inducer:
Fair enough. Thanks for helping to contain the messiness of the IPython compat code.

asmeurer:
As long as we still support IPython 4 I'm happy.

PR #213, opened 2017-01-05
Check that we're on the main thread before installing the signal handler

### LINKED ISSUE #212: Trying to interrupt a different thread prints a warning
Tested with Python 3.5.2 and 2.7.12

Sample program:

```python
import threading


class Foo(threading.Thread):
  def run(self):
    import pudb; pu.db

Foo().start()
```

pudb version: 2016.2

Error:

```
/pudb/__init__.py:201: UserWarning: setting interrupt handler on signal 2 failed: Traceback (most recent call last):
  File "/tmp/foo/env/local/lib/python2.7/site-packages/pudb/__init__.py", line 195, in set_interrupt_handler
    signal.signal(interrupt_signal, _interrupt_handler)
ValueError: signal only works in main thread
```

signal.signal only working in the main thread is well documented (https://docs.python.org/2/library/signal.html#signal.signal). We can detect if the current thread is the main thread and just skip setting the signal handler (`isinstance(threading.current_thread(), threading._MainThread)` - from http://stackoverflow.com/a/23207116), but I don't know what functionality we loose if we do that.

cdman:
Makes the error message a little bit more clear and also surpresses setting the signal handler when not on the main thread and not explicitly asked to install one.

Closes #212.

### DIFF
--- a/pudb/__init__.py
+++ b/pudb/__init__.py
@@ -18,7 +18,9 @@ class PudbShortcuts(object):
         import sys
         dbg = _get_debugger()
 
-        set_interrupt_handler()
+        import threading
+        if isinstance(threading.current_thread(), threading._MainThread):
+            set_interrupt_handler()
         dbg.set_trace(sys._getframe().f_back)
 
 if PY3:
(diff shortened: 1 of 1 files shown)


inducer:
Thanks for your contribution!

PR #214, opened 2017-01-09
Display None values as 'None' in the variable viewer

Wilfred:
Previously they were 'NoneType', but this is a safe value to display.


inducer:
Thanks for your contribution!

Issue #221, opened 2017-02-03
Provide an interface consistent with pdb so that pudb can be used by pytest


anntzer:
pytest provides the switch `--pdb` in order to enter a debugger whenever a test fails.  In https://github.com/pytest-dev/pytest/pull/1712 I added the `--pdbcls` flag to pytest so that one can specify an arbitrary `Pdb` subclass; e.g. one can call `py.test --pdbcls=IPython.core.debugger:Pdb --pdb` to use the ipython debugger instead.

Unfortunately it is currently not possible to pass in pudb (`py.test --pdbcls=pudb.debugger:Debugger --pdb`) because the `Debugger.interaction` method has a completely different signature than `Pdb.interaction` (which is the entry point used), leading to a crash of pytest.  Specifically, the signatures are
```
pdb.Pdb.interaction(self, frame, traceback)
```
and
```
pudb.debugger.Debugger.interaction(self, frame, exc_tuple=None, show_exc_dialog=True)
```

Would it be possible to modify the signature of `Debugger.interaction` so that the second argument becomes, say, `exc_tuple_or_traceback` (to maintain backcompatibility and also be compatible with pytest)?

asmeurer:
It sounds good to me. Have you tested it to see if there are any other compatibility issues that pop up?

anntzer:
Well, I haven't implemented anything and am not so familiar with pudb's internals so I can't tell.

inducer:
Likewise, sounds good to me. Try it out and send a PR if it turns out to work.

PR #224, opened 2017-02-20
Make Debugger.interaction match the signature of Pdb.interaction.

anntzer:
This allows pudb to be used as debugger for pytest>=3.0, e.g.

```
$ cat test_foo.py
def test_1():
    assert False
$ pytest --pdbcls pudb.debugger:Debugger --pdb --capture=no
```
(starts pudb at the failed assertion site).

Note the need to pass `--capture=no` (or its synonym `-s`) as otherwise
pytest tries to manage the standard streams itself.

Backwards compatibility is preserved.

See #221.

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -4,8 +4,10 @@
 from __future__ import absolute_import, division, print_function
 import urwid
 import bdb
-import sys
+import gc
 import os
+import sys
+from types import TracebackType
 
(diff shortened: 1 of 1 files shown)


inducer:
Thanks!

inducer:
I'll add a line to the README to make sure people know about how to run pytest.

PR #227, opened 2017-03-14
fix bug compile failed in Python 2.X

hehaha:
Toggling breakpoint and Running to cursor will be failed if there is "print 'xxx'" in the code. Because the print_function mock the 'print' in the code. When you call at the get_executable_lines_for_file function and compile the file, it will generate a syntax error.

### DIFF
--- a/pudb/lowlevel.py
+++ b/pudb/lowlevel.py
@@ -1,4 +1,4 @@
-from __future__ import absolute_import, division, print_function
+from __future__ import absolute_import, division
 from pudb.py3compat import PY3
 
 


inducer:
Thx!

asmeurer:
It would be better to set the `flags` argument to `compile`. The `__future__` import could easily be added back in accidentally.

hehaha:
It's indeed a better solution! I pulled another [request](https://github.com/inducer/pudb/pull/228) to fix it. I didn't understand the flags argument, but I read the document and found if I set the dont_inherit to a non-zero integer and the compile function would ignore the  future statements. So this is what I did.

PR #235, opened 2017-04-03
Add set_continue() function to __init__.py

### LINKED ISSUE #234: Allow starting without pause
I want to make PuDB available in my script, but when I do `pudb.set_trace()`, I don't want it to stop there. I want it to stop only on breakpoints that I've already set (Saved in that config file.)

I'm not seeing a way to do that with PuDB, if there is maybe it should be in the readme.

asmeurer:
This lets you start the debugger and immediately continue, i.e., it only stops
on a previously defined breakpoint.

Fixes #234.

Let me know if you would prefer a better name, or just a flag to set_trace().

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -118,24 +118,25 @@ PuDB also has a `mailing list <http://lists.tiker.net/listinfo/pudb>`_ that
 you may use to submit patches and requests for help.  You can also send a pull
 request to the `GitHub repository <https://github.com/inducer/pudb>`_
 
-Attaching to Running Code
--------------------------
+Starting the debugger without breaking
+--------------------------------------
 
-An alternative to using ``set_trace`` is to use::
+To start the debugger without actually pausing use::
 
-    from pudb import activate_and_continue; activate_and_continue()
+    from pudb import set_trace; set_trace(paused=False)
 
 at the top of your code.  This will start the debugger without breaking, and
-run it until a predefined breakpoint is hit.
+run it until a predefined breakpoint is hit. You can also press ``b`` on a
+``set_trace`` call inside the debugger, and it will prevent it from stopping
+there.
 
 Interrupt Handlers
 ------------------
 
-Both ``set_trace`` and ``activate_and_continue`` set ``SIGINT`` (i.e.,
-``Ctrl-c``) to run ``set_trace``, so that typing ``Ctrl-c`` while your code is
-running will break the code and start debugging. See the docstring of
-``set_interrupt_handler`` for more information. Note that this only works in
-the main thread.
+``set_trace`` sets ``SIGINT`` (i.e., ``Ctrl-c``) to run ``set_trace``, so that
+typing ``Ctrl-c`` while your code is running will break the code and start
+debugging. See the docstring of ``set_interrupt_handler`` for more
+information. Note that this only works in the main thread.
 
 Programming PuDB
 ----------------
--- a/README.rst
+++ b/README.rst
@@ -123,12 +123,19 @@ Attaching to Running Code
 
 An alternative to using ``set_trace`` is to use::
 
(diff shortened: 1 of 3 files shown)


inducer:
`pudb.activate_and_continue()` and perhaps `pu.cont`?

inducer:
Also needs a shout-out in the README.

inducer:
Lastly, thanks for working on this!

asmeurer:
How about replacing the `set_interrupt_handler` bit in the README with this, since it also sets the handler?

asmeurer:
Still not thinking `activate_and_continue` is the best name. Think maybe it should just be a keyword argument to set_trace. I'll let you decide, though.

inducer:
> Still not thinking activate_and_continue is the best name.

Got a better one?

>Think maybe it should just be a keyword argument to set_trace.

`set_trace` evokes breakpoint to me, just because that's it's traditional use in `pdb`. Not sure I like this much.

cool-RR:
I'm okay with whatever name you choose, or a keyword argument to `set_trace`. As long as it works ;)

If you'd ask me what the best naming is... I would say that `set_trace` is a stupid name to begin with. Yes, I get it, it's named that way because debuggers use `sys.settrace`. But who cares. Users shouldn't be concerned with how your program is implemented, they care about what it can do for them. So any attempt to make naming intuitive would start from providing an alias to `set_trace`, maybe something like `start_debugger`, and then putting keyword arguments on that like `stop_at_start=True` would be the best thing. 

Of course, this is would be a big change and that's just my opinion :) So I completely accepted the fact that it's not likely to ever happen.

inducer:
`pudb.start(continue=False)`?

cool-RR:
I'm okay with the start name. The continue name is both unclear and illegal
:)

Sent from my phone.

On Apr 3, 2017 18:47, "Andreas Klöckner" <notifications@github.com> wrote:

pudb.start(continue=False)?

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/inducer/pudb/pull/235#issuecomment-291184291>, or mute
the thread

jeffwidman:
Thanks for documenting this clearly, I appreciate it.

inducer:
`start(interactive=True)`?

cool-RR:
Definitely not. It's misleading.

start_paused sounds best to me. Can be start_stopped or pause_on_start or
paused. Just my opinion though.

inducer:
Gpod names are hard. I appreciate the feedback!

inducer:
`start(paused=True)`?

`start(keep_going=False)`?

cool-RR:
`start(paused=True)` is great.

inducer:
OK, looks like we've got a winner. @asmeurer, do you agree?

asmeurer:
Sounds fine to me. `start` is an alias for `set_trace` I take it.

cool-RR:
@asmeurer Thank you very much for implementing this so quickly!

Are you going to make a release for these features?

asmeurer:
OK, take a look. Making it a keyword argument simplifies the code quite a bit. Although when testing it I admit `paused=False` confused me more than once because it's negated (False). I really wish it could be `continue=True` (`_continue=True`?). `stop=False` is another suggestion. 

I also noticed another, mostly harmless, bug where if you save a breakpoint on a set_trace() line, you have to disable it twice to disable it, once for the saved breakpoint and once for the magic breakpoint (this would show up if you pressed 'b' on a set_trace(paused=False) to make it start breaking there, and then later removed the pause from the call and wanted to unbreakpoint it interactively).

asmeurer:
@inducer handles the releases, but he usually does so once pull requests are merged.

inducer:
`pu.cont` should probably be renamed, too. `pu.go`?

inducer:
Thx!

PR #236, opened 2017-04-04
Don't use SetPropagatingDict in the built in console

### LINKED ISSUE #166: Ctrl-x not function to leave console
When attempting to make local calls in the console, pressing ctrl-x merely prints `^x` rather than leaving the console. At this point, the only way out is to kill pudb. An interesting behavior is that this only happens after issuing a command in the console; ctrl-x works fine until then. 

Details: 
pudb version: 2015.4.1
installed with pip
running in osx 10.11.2 in iTerm2, Terminal, and Neovim's terminal mode

asmeurer:
Passing a dict subclass to eval is not supported in Python 2. This was causing
weird issues where the C-x would break the UI (I am not clear exactly how it
caused those issues).

The point of SetPropagatingDict is to combine globals and locals into a single
dictionary for things like code.InteractiveConsole and rlcompleter.Completer,
which take a single dictionary, but to allow changes to propagate, so that
assignments in the shell can modify global names. However, this is completely
unnecessary for the built in shell, as we call eval() ourselves, so we can
just pass the exact globals and locals dicts separately.

This fixes #166.

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -1554,8 +1554,10 @@ class DebuggerUI(FrameVarInfoKeeper):
             sys.stdin = None
             sys.stderr = sys.stdout = StringIO()
             try:
+                # Don't use cmdline_get_namespace() here, it breaks things in
+                # Python 2 (issue #166).
                 eval(compile(cmd, "<pudb command line>", 'single'),
-                        cmdline_get_namespace())
+                        self.debugger.curframe.f_globals, self.debugger.curframe.f_locals)
             except:
                 tp, val, tb = sys.exc_info()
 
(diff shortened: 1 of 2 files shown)


asmeurer:
I'll also add some comments on SetPropagatingDict so that we can know what it is for in the future.

inducer:
> I'll also add some comments on SetPropagatingDict so that we can know what it is for in the future.

Awesome, thx!

asmeurer:
Another point of confusion. The usage everywhere is `SetPropagatingDict([locals, globals], locals)`, meaning assignments are supposed to go to `locals`. But assigning to `locals` doesn't work. The only reason it works is that at the module level, `locals() is globals()`. I suppose this is the desired behavior, though, assignments in the shell should work if and only if we are at the module level. It definitely needs some clarification in the code, however.

inducer:
Thx!

Issue #115, opened 2014-04-21
has trouble tracing importlib._bootstrap in python3.3


bukzor:
I'm doing an experiment with sys.path_hooks, and pudb is having trouble tracing through it.

pudb seems to fail to find the source for `/usr/lib/python3.3/importlib/_bootstrap.py`, which results in the variables windowlet being smashed to the left.

To reproduce:

``` sh
$ PYTHONPATH=dummypath python3.3 foo.py
```

`foo.py`:

``` python
from __future__ import print_function


class NoPathHook(object):
    def __init__(self, syspath):
        if syspath.endswith('/dummypath'):
            pass
        else:
            raise ImportError

    @staticmethod
    def find_module(module):
        if module == 'example_thingy_doesnt_exist':
            import pudb.b


def register():
    import sys
    sys.path_hooks.insert(0, NoPathHook)
    sys.path_importer_cache.clear()


def main():
    register()
    try:
        import example_thingy_doesnt_exist
    except ImportError as error:
        if error.name == 'example_thingy_doesnt_exist':
            pass
        else:
            raise

    print('DONE')

if __name__ == '__main__':
    exit(main())
```

wronglink:
Seems to be fixed at #240

PR #240, opened 2017-04-05
Add info about linecache to the NullSourceCodeProvider message

### LINKED ISSUE #238: Source code not found should recommend adding lines to linecache
If the source code is not found, it says 

```
If this is generated code and you would like the source code to show up here,
simply set the attribute _MODULE_SOURCE_CODE in the module in which this function
was compiled to a string containing the code.
```

I think it should also recommend adding the lines to `linecache`, which is doable even if you don't control the module, and will add support for other things as well (like the `traceback` module and `inspect.getsource`).

asmeurer:
Fixes #238.


inducer:
Thx!

wronglink:
Seems to me it also fixed #115

Issue #242, opened 2017-04-05
Increase the code coverage and PEPs standards


discort:
**Problem**
There are no tests so developers can't make pull requests without breaking the code. The issue probably related with #40 
Also, need to standardize the code as much as possible.
At the moment these PEPs are ignored: `E126,E127,E128,E123,E226,E241,E242,E265,W503,E402`

**Solution**
1. Increase the code coverage as much as possible.
2. Integrate with CI, (i.e. CircleCI, TravisCI). Add a test runner for the each pull request.
3. Step by step make a code refactoring and standardize simultaneously, later add linter for PRs.

asmeurer:
I'm not too concerned with PEP 8 things. Most of the rules you cited are things I don't completely agree with, and definitely don't think should be enforced with tests. [Many](https://pycodestyle.readthedocs.io/en/latest/intro.html#error-codes) are so controversial that they aren't even enabled by default in pycodestyle (previously called pep8). 

I do agree that we need tests (it's mostly an issue of figuring out how, see https://github.com/inducer/pudb/issues/40). Running pyflakes could be helpful too (unlike pycodestyle, pyflakes only flags logical errors, not style errors). It would need to be configured to ignore Python 3 compatibility stuff and some other false positives.

discort:
The main issue is testing not PEP8 things (PEP 8 just helps different people to use the similar code style).

> I do agree that we need tests (it's mostly an issue of figuring out how, see #40).
Start from a testing of single functions and increase the code coverage, then do more complicated stuff

PR #243, opened 2017-04-05
Added a base tests for settings

discort:
#242 
Execute `pip install -r requirements.dev.txt`

To run tests with a coverage:
`py.test --cov=pudb tests`

Tested with python3.5 and 2.7


asmeurer:
I didn't even realize that we already had some tests. We should enable Travis or Circle (I have no preference), so that we can at least see that they are working.

inducer:
FWIW, I already have CI running on my internal gitlab: https://gitlab.tiker.net/inducer/pudb

inducer:
(But I'm not against switching to Travis/Circle.) Config is in `.gitlab-ci.yml`.

discort:
What about to add the information to a docs
1. How to setup a development environment
2. How to run tests locally

inducer:
Added both of you to the Gitlab project. You should have received email--check your spam folders.

inducer:
@discort It'd be fantastic if you could move this pull request there to have the CI run on it.

discort:
I have no idea how to use GitLab. I pushed my branch there but I suppose I did it wrong

inducer:
You're totally fine. Created a merge request for you: https://gitlab.tiker.net/inducer/pudb/merge_requests/1

asmeurer:
Is it possible to have GitLab run the tests on pull requests automatically, and update the GitHub status?

asmeurer:
I see, so it's already running flake8, with those given things ignored. So you just need to add ignore flags for the unused imports in the compat file.

asmeurer:
Also, flake8 ought to be run in each Python build, not as a separate build. That way, it can catch Python 2/3 incompatibilities (e.g., if you use `unicode` in Python 2 without importing it from compat, flake8 will fail if run in Python 3 but not in Python 2).

inducer:
> Also, flake8 ought to be run in each Python build, not as a separate build.

It'll stay a  separate build, but if you'd like to run it for 2 and 3, feel free to hack this: https://gitlab.tiker.net/inducer/ci-support/

PR #245, opened 2017-04-06
Added TravisCI and codecov.io support

discort:
1. Register at https://travis-ci.org and here https://codecov.io
2. Syncronize `pudb` repo.
3. In the TravisCI setting set environment variable `COVERAGE` with a value from of the upload token from codecov.io (see pudb settings at codecov.io)
4. The badges badges will be added to readme.rst

An example: https://github.com/discort/pudb

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,25 @@
+[run]
+branch = True
+
+[report]
+# Regexes for lines to exclude from consideration
+exclude_lines =
+    # Have to re-enable the standard pragma
+    pragma: no cover
+
+    # Don't complain about missing debug-only code:
+    def __repr__
+    if self\.debug
+
+    # Don't complain if tests don't hit defensive assertion code:
+    raise AssertionError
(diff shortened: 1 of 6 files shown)


inducer:
Thx!

discort:
@inducer 
Need to add `__init__.py` to a `test` folder

inducer:
FYI: Passes now. Thanks for setting this up.

inducer:
Coverage isn't working. I've added `CODECOV_TOKEN` with the correct value to the Travis env vars, but... nothing.

inducer:
Maybe the upload script is looking in the wrong spot--could you take a look?

discort:
@inducer 
You added `CODECOV_TOKEN` env variable but script uses `COVERAGE` https://github.com/inducer/pudb/blob/master/upload_coverage.sh#L3

These names should be the same

inducer:
Fixed, as of a while ago. Thanks for your help.

PR #247, opened 2017-04-11
Miscellaneous code cleanups

wronglink:
Hi. I'd like to propose and discuss some miscellaneous codestyle changes, that I think improve code readability and ease other contributors to work with. I've split up changes into commits, to be able to apply them separately. So I don't mind to update PR if some of changes are not approved.

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -541,10 +541,7 @@ class FileSourceCodeProvider(SourceCodeProvider):
         self.file_name = debugger.canonic(file_name)
 
     def __eq__(self, other):
-        return (
-                type(self) == type(other)
-                and
-                self.file_name == other.file_name)
+        return type(self) == type(other) and self.file_name == other.file_name
 
     def identifier(self):
         return self.file_name
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -7,6 +7,7 @@ import bdb
 import gc
 import os
 import sys
+from functools import partial
 from types import TracebackType
 
 from pudb.settings import load_config, save_config
@@ -19,18 +20,6 @@ if PY3:
 else:
     _next = "next"
 
-try:
-    from functools import partial
-except ImportError:
-    def partial(func, *args, **keywords):
-        def newfunc(*fargs, **fkeywords):
-            newkeywords = keywords.copy()
-            newkeywords.update(fkeywords)
-            return func(*(args + fargs), **newkeywords)
-        newfunc.func = func
-        newfunc.args = args
-        newfunc.keywords = keywords
-        return newfunc
 
 HELP_TEXT = """\
 Welcome to PuDB, the Python Urwid debugger.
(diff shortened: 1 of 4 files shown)


asmeurer:
Looks fine to me. I believe everything you added is supported by Python 2.6, which would be my main concern.

discort:
Ideally, you should cover your code by tests. I created a PR how to start that
https://github.com/wronglink/pudb/pull/1

@wronglink

wronglink:
@discort thanks for idea. I'll provide tests for this PR.

@asmeurer I tried to use only 2-3 portable technics, but with tests we would have less doubts about it.

wronglink:
Added more tests. Think I'm done with it.

test/test_lowlevel.py
```diff
@@ -0,0 +1,31 @@
+# -*- coding: utf-8 -*-
+from pudb.lowlevel import detect_encoding, decode_lines
+from pudb.py3compat import PY3
+
+
+def test_detect_encoding_nocookie():
+    lines = ['Test Проверка']
```
wronglink: Not sure, but I don't think so. On py3 they would be unicodes, on py2 strings. As I can see, `detect_encoding` works with native string format (unicode for 3, str for 2). So it would when receive unicode lines on 2 and bytes on 3.

asmeurer: It's been awhile since I've dealt with this stuff so I can't remember if it is needed. The `u` does nothing in Python 3, so it would only be a question of the Python 2 behavior. I had thought that strings without `u` in Python 2 with non-ASCII characters are undefined, but perhaps the coding line at the top of this file fixes that. I'm also not clear, but if this is the specific thing that this function is testing, then this is fine.

wronglink: For now I think it's quite OK to cover more code with tests to kind of freeze current codebase stage.

wronglink:
@inducer I think you review is needed here.

pudb/debugger.py
```diff
@@ -612,12 +584,11 @@ def __init__(self, func_name, code):
...
-                type(self) == type(other)
-                and
-                self.function_name == other.function_name
-                and
-                self.code is other.code)
+        return all([
+            type(self) == type(other),
+            self.function_name == other.function_name,
+            self.code is other.code,
+        ])
```
inducer: This is weird to me. It incurs an extra function call penalty *and* it's less readable (to me).

pudb/debugger.py
```diff
@@ -1875,20 +1832,20 @@ def help(w, size, key):
...
-                and CursesScreen is not None):
+        want_curses_display = any([
+            CONFIG["display"] == "curses",
+            all([
+                CONFIG["display"] == "auto",
+                not any([
+                    os.environ.get("TERM", "").startswith("xterm"),
+                    os.environ.get("TERM", "").startswith("rxvt")
+                ])
+            ])
```
inducer: Same.

pudb/debugger.py
wronglink: OK. I've reverted these kind of cleanups.

inducer:
Thanks!

PR #249, opened 2017-04-13
Add support for custom shells

asmeurer:
This add support for custom shells, similar to custom stringifiers. 

A custom shell can be defined by creating a file with a function `pudb_shell(_globals, _locals, first_time)` at the module level, and setting it in the settings.

The `first_time` setting is there to match the existing shells, but it seems useless (a function can easily tell if it has already been called), so it should perhaps be removed.

I also added some documentation for shells in general.

### DIFF
--- a/example-shell.py
+++ b/example-shell.py
@@ -46,3 +46,4 @@ def pudb_shell(_globals, _locals):
     from code import InteractiveConsole
     cons = InteractiveConsole(ns)
     cons.interact("Press Ctrl-D to return to the debugger")
+    # When the function returns, control will be returned to the debugger.
--- a/example-shell.py
+++ b/example-shell.py
@@ -4,7 +4,7 @@ shell used when pressing the ! key in the debugger (it does not affect the
 Ctrl-x shell that is built into PuDB).
 
 To create a custom shell, create a file like this one with a function called
-pudb_shell(_globals, _locals, first_time) defined at the module level. Note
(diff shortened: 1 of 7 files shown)


example-shell.py
```diff
@@ -0,0 +1,55 @@
...
+
+The example in this file
+
+"""
+
+# Define this a function with this name and signature at the module level.
+def pudb_shell(_globals, _locals, first_time):
+    """
+    This example shell runs a classic Python shell. It is based on
+    run_classic_shell in pudb.shell.
```
asmeurer: Let me know if you can think of a better example.

asmeurer:
I think I got `locals` and `globals` backwards in a few places (apparently PuDB uses the reverse convention as `exec`). There's no issue with breaking the API of functions in shell.py, right?

asmeurer:
Well I went ahead and change the API. This is ready for review.

inducer:
LGTM, but has picked up some conflicts. Out of curiosity, what's the use case? How many more shells are there that we don't yet cover...?

asmeurer:
The use case is that I wrote my [own shell](https://github.com/asmeurer/mypython). However, I have no plans to distribute it widely. 

That's probably a rare situation, but I suppose this could also be used to customise the shell used by PuDB (e.g., IPython with some custom settings). Or maybe someone would want something crazy like opening the notebook.

inducer:
LGTM, thx!

PR #256, opened 2017-05-18
Stop swallowing all exceptions in settings.py

### LINKED ISSUE #139: "Edit preferences" comes up every time I use pdb
I assume there is a way to fix this, but I haven't found this after 15 minutes. It would be great if the Edit Preferences page told you how to suppress it on the next pudb launch.

jeffwidman:
Stop swallowing all exceptions in `settings.py`. 

This is a broader swath than the original issue which only discussed removing the bare `try`/`except` from `load_config()`. However, I can envision the same type of problem happening in `save_config()` resulting in someone being confused why their configs aren't being saved. 

Likely we'll need to explicitly catch a handful of exceptions, but at this point we're not sure which ones they are.

I'm also very open to reworking this to only remove the one `try`/`except` clause, but I figured might as well start the conversation by trying to fix all the things... 😄 

Fix #139

### DIFF
--- a/pudb/settings.py
+++ b/pudb/settings.py
@@ -51,15 +51,12 @@ def load_config():
     cparser = ConfigParser()
 
     conf_dict = {}
-    try:
-        cparser.read([
-            join(cdir, XDG_CONF_RESOURCE, CONF_FILE_NAME)
-            for cdir in XDG_CONFIG_DIRS if isdir(cdir)])
+    cparser.read([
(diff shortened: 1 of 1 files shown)


inducer:
Thanks for your contribution! The tests failed--could you take a look?

jeffwidman:
Yeah, I saw that, probably won't have time to fix until next week.

mvanderkamp:
I think this can be closed- the exceptions are being logged now.

inducer:
Agreed, thanks!

PR #259, opened 2017-06-25
Tell ptpython to read/process the user's config file.

jsivak:
Am using ptpython as the embedded shell during a pudb debugging session and noticed that ptpython was not reading my settings stored in ~/.ptpython/conf.py.

Based on the info from https://github.com/django-extensions/django-extensions/pull/716/commits/36fa486d23206a766c3507c36b2bd767e190e98e and https://github.com/django-extensions/django-extensions/issues/711 I was able to update the pudb code to instruct ptpython to read the users config file.

Let me know if you have any questions.

Thanks

### DIFF
--- a/pudb/shell.py
+++ b/pudb/shell.py
@@ -12,7 +12,7 @@ else:
     HAVE_BPYTHON = True
 
 try:
-    from ptpython.repl import embed as ptpython_embed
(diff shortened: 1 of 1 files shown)


inducer:
Thanks!

PR #264, opened 2017-08-24
Don't show "Return: None" when "returning" from a module frame

asmeurer:
That is, when the last line of a module file is executed (usually from
runscript).


inducer:
LGTM--feel free to merge once the tests pass.

asmeurer:
Must be some bug. The tests did pass.

asmeurer:
Hmm, now with this change, if the last line in a file is a variable definition, pudb doesn't show it in the variable view.

PR #268, opened 2017-08-29
Fix keyword arg error

Alok:
`embed` expects `globals` and `locals` to be passed as keyword
arguments, and this fixes it.


inducer:
Thanks!

PR #272, opened 2017-09-03
Don't return early when "returning" from a module

asmeurer:
We still need to run interaction so that the debugger will show the result of
the last line that was run. We just don't want it to treat the end of the
module as a "return".

For instance, if you have a file with a single variable definition

    a = 1

then if you stop on the definition, pudb will show the variable "a: 1" in the
variables view (but it won't show "return: None" after stepping again).

Fixes https://github.com/inducer/pudb/pull/264#issuecomment-326777042.

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -382,10 +382,8 @@ class Debugger(bdb.Bdb):
 
     def user_return(self, frame, return_value):
         """This function is called when a return trap is set here."""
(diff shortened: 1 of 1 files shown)


asmeurer:
Is it possible to add a test for this?

inducer:
> Is it possible to add a test for this?

I can't think of a straightforward way...

At any rate, thanks for working on this!

Issue #271, opened 2017-09-02
Latest PuDB (2017.1.3) dies with ListBoxError in some cases


cdman:
Stacktrace (edited slightly for anonymity):

```
  File "/usr/lib/python2.7/bdb.py", line 49, in trace_dispatch
    return self.dispatch_line(frame)
  File ".../site-packages/pudb/debugger.py", line 160, in dispatch_line
    self.user_line(frame)
  File ".../site-packages/pudb/debugger.py", line 381, in user_line
    self.interaction(frame)
  File ".../site-packages/pudb/debugger.py", line 349, in interaction
    show_exc_dialog=show_exc_dialog)
  File ".../site-packages/pudb/debugger.py", line 2084, in call_with_ui
    return f(*args, **kwargs)
  File ".../site-packages/pudb/debugger.py", line 2322, in interaction
    self.event_loop()
  File ".../site-packages/pudb/debugger.py", line 2280, in event_loop
    canvas = toplevel.render(self.size, focus=True)
  File ".../site-packages/urwid/widget.py", line 141, in cached_render
    canv = fn(self, size, focus=focus)
  File ".../site-packages/urwid/widget.py", line 1751, in render
    canv = get_delegate(self).render(size, focus=focus)
  File ".../site-packages/urwid/widget.py", line 141, in cached_render
    canv = fn(self, size, focus=focus)
  File ".../site-packages/urwid/container.py", line 1083, in render
    focus and self.focus_part == 'body')
  File ".../site-packages/urwid/widget.py", line 141, in cached_render
    canv = fn(self, size, focus=focus)
  File ".../site-packages/urwid/decoration.py", line 225, in render
    canv = self._original_widget.render(size, focus=focus)
  File ".../site-packages/urwid/widget.py", line 141, in cached_render
    canv = fn(self, size, focus=focus)
  File ".../site-packages/urwid/container.py", line 2085, in render
    focus = focus and self.focus_position == i)
  File ".../site-packages/urwid/widget.py", line 141, in cached_render
    canv = fn(self, size, focus=focus)
  File ".../site-packages/urwid/widget.py", line 1751, in render
    canv = get_delegate(self).render(size, focus=focus)
  File ".../site-packages/urwid/widget.py", line 141, in cached_render
    canv = fn(self, size, focus=focus)
  File ".../site-packages/urwid/container.py", line 1526, in render
    canv = w.render((maxcol, rows), focus=focus and item_focus)
[...]

PR #273, opened 2017-09-03
Improve some text width calculations

asmeurer:
Preliminary fix for the issues described at https://github.com/inducer/pudb/issues/271. Still need confirmation that it fixes https://github.com/inducer/pudb/issues/271 (I don't know that it does). And I still need to add a lot of tests. We need to better hammer this code in the tests so that all the double- and zero-width corner cases don't crash pudb.

### DIFF
--- a/pudb/ui_tools.py
+++ b/pudb/ui_tools.py
@@ -5,6 +5,16 @@ from urwid.util import _target_encoding, calc_width
 
 # generic urwid helpers -------------------------------------------------------
 
+def text_width(txt):
+    """
+    Return the width of the text in the terminal
+
+    Use this instead of len() whenever txt could contain double- or zero-width
+    Unicode characters.
+
+    """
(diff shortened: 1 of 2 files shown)


inducer:
This looks good to me. If you'd like to merge it now, I say go ahead. Alternatively, if you'd like to wait until there are more tests or confirmation from @cdman, that's fine, too.

asmeurer:
I want to write some tests first.

asmeurer:
I'm also unsure if my changes to var_view are needed. I haven't yet found a bug that actually is fixed by them.

inducer:
As evidenced by #271, I suspect we're better off merging this than not. The tests can always go in later. What do you think? You have my go-ahead to merge this.

asmeurer:
I tested and the changes in var_view aren't what fixed that issue. But I agree we should merge it.

Issue #276, opened 2017-10-01
How do I prevent the 'safely_stringify_for_pudb' attribute tests when inspecting code?


emehrkay:
Python 3.6 (2.7 is okay)

My object is setup to capture all `__getattr__` requests and build other objects from it. So when I am inspecting my code, pudb is always running this check and messing things up. 

The only solution that I could come up with was to add a method like this:

```python
    def safely_stringify_for_pudb(self):
        return None
```

There is a better way right?

inducer:
I'd be happy to take a patch that either makes a global setting to turn this off or that turns this off for objects that override `__getattr__`.

emehrkay:
It would probably make more sense as a global setting. I'll take a look through the code and see if I can come up with anything.

offero:
This attribute causes an exception in my code that uses suds, so I can't use pudb with it.

inducer:
I'd be happy to consider a patch that makes your life better.

lega911:
version 2017.1.2 works well
```pip3 install -U --user pudb==2017.1.2```

inducer:
How is 2018.1 (just released)? (with #276)

lega911:
2018.1 seems works fine, tested on python 3.6

PR #281, opened 2017-10-15
check 'safely_stringify_for_pudb' on the type #276

koreno:
The hasattr call should check for 'safely_stringify_for_pudb' on the type of the object
instead of the object itself, so as not to implicitly call its __getattr__ (if implemented).
This will avoid the exceptions on such magic objects.


inducer:
Good idea, thanks!

PR #283, opened 2017-11-27
set socket REUSEADDR opt for the remote debugger

lins05:
Otherwise sometimes if the remote debugger would use 6900 even after the program is restarted.


inducer:
Thanks! Learned something about sockets as part of this.

PR #287, opened 2017-12-28
Show length of array types in var view

Allda:
The new code shows length of set, frozenset, list, tuple and dict type
varibles in variable view.

### DIFF
--- a/pudb/var_view.py
+++ b/pudb/var_view.py
@@ -238,6 +238,9 @@ def type_stringifier(value):
         except Exception:
             pass
 
+    elif type(value) in [set, frozenset, list, tuple, dict]:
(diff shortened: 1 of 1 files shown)


inducer:
@asmeurer, what do you think? My only thought is that, depending on the stringifier, this information could be somewhat redundant. But I could also see it being helpful. I'm +0.

asmeurer:
I would do it just for the Type (default) stringifier.

inducer:
Agreed. @Allda, could you move that code into that stringifier?

Allda:
@asmeurer @inducer I moved the code to `type_stringifier()` function

inducer:
LGTM. Thanks!

PR #288, opened 2018-01-08
Use `isinstance()` rather than `type()`

jeffwidman:
Use `isinstance()` rather than `type()` so that we also match against subclasses of these types.

### DIFF
--- a/pudb/var_view.py
+++ b/pudb/var_view.py
@@ -238,7 +238,7 @@ def type_stringifier(value):
         except Exception:
             pass
 
-    elif type(value) in [set, frozenset, list, tuple, dict]:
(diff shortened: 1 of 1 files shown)


inducer:
No, that's intentional. The issue is that `__len__` may not be safe to call on a subclass as it hits user code. (Though maybe that's being too paranoid... what do you think?)

asmeurer:
Yeah this is a case where a direct type check actually is correct.

jeffwidman:
Good point. I can't think of a clean way to check whether `__len__` is safe to call, as I think the hassle of calling it, verifying it returns an integer and swallowing any exceptions seems like more hassle than the benefit.

PR #293, opened 2018-04-10
Show variables that start with an underscore in the var view

### LINKED ISSUE #292: variables that start with underscores or dots not shown
From this line https://github.com/inducer/pudb/blob/3f627ce0f7370ab80bc2496cb3d2364686f10efe/pudb/var_view.py#L556 variables that start with `_` or `.` in the current namespace aren't show. This is separate from the method name filtering. For instance

```
a = 1
_b = 2
c = 3
```

if you run pudb on this script, only `a` and `b` are shown. 

Why is this filter here? It seems to me it should only filter things that start and end with `__`, to get rid of `__builtins__`, `__file__`, and so on. And why is the filter for `.` there? A variable name can only start with `.` if something manually put it in `globals()`.

asmeurer:
We still filter things that start and end with `__`, to filter out things like
`__builtins__`, `__file__`, and so on. It also removes the filtering of variables
that start with `.`, which is only possible if something puts that in globals()
manually.

Fixes #292.

### DIFF
--- a/pudb/var_view.py
+++ b/pudb/var_view.py
@@ -553,7 +553,7 @@ def make_var_view(frame_var_info, locals, globals):
                 attr_prefix="return")
 
     for var in vars:
-        if not var[0] in "_.":
+        if not (var.startswith('__') and var.endswith('__')):
             tmv_walker.walk_value("", var, locals[var])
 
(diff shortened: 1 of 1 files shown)


inducer:
Thx!

PR #302, opened 2018-06-16
Use "space" as additional key for expanding/collapsing variables

eht16:
Space is on some keyboard layouts (like German) way easier to reach than
backslash, so add "space" as an alternative key for expanding/collapsing
variables.


asmeurer:
Sounds fine to me.

eht16:
I just realized it might be a good idea to add the new shortcut to the help text :). Done in 594bd02.

inducer:
LGTM. Thanks for your contribution!

PR #303, opened 2018-06-16
Add setting to make access level in variables view persistent

eht16:
Currently, all variables in the var view have the default access level of "public". This might be a sane default but users may wish to change their default to something else. E.g. I prefer to have "private" as default and so add a setting for the access level to be able to change the default.

### DIFF
--- a/pudb/settings.py
+++ b/pudb/settings.py
@@ -106,6 +106,7 @@ def load_config():
     conf_dict.setdefault("custom_shell", "")
 
     conf_dict.setdefault("wrap_variables", True)
+    conf_dict.setdefault("variables_access_level", "public")
 
     conf_dict.setdefault("display", "auto")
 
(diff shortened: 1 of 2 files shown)


inducer:
Thanks for your contribution! One possible concern I have with this is that it doesn't conform to the "instant-apply" semantics used by the other settings, since this setting is only considered when new variable views are created.

We may decide that those semantics are just fine, but I am interested in at least a brief discussion.

inducer:
@asmeurer what's your take/

pudb/settings.py
```diff
@@ -213,6 +217,13 @@ def _update_config(check_box, new_state, option_newvalue):
 
                 conf_dict.update(stringifier=newvalue)
                 _update_stringifier()
+
+        elif option == "variables_access_level":
+            # only activate if the new state of the radio button is 'on'
+            if new_state:
+                conf_dict.update(variables_access_level=newvalue)
```
inducer: I think I'd prefer if these could be renamed `default_variables_access_level`.

eht16: Done. I renamed all occurrences to keep it consistent.

inducer:
Thanks for your contribution, and for being responsive to comments. LGTM.

eht16:
You're welcome. Thanks for this awesome debugger!

Issue #129, opened 2015-01-07
Jump command


asmeurer:
When researching the bug for https://github.com/inducer/pudb/pull/90, I read in the [docs](https://docs.python.org/3/reference/datamodel.html):

> f_lineno is the current line number of the frame — writing to this from within a trace function jumps to the given line (only for the bottom-most frame). A debugger can implement a Jump command (aka Set Next Statement) by writing to f_lineno.

I'm probably the last person to learn about this, but it seems like a useful thing to implement. `j` is already taken as down, but `J`. It's probably easy to implement, at least assuming the error reporting is good for bad jumps.

inducer:
I wouldn't turn this down if there were a patch, although IMO it's not exactly mind-blowing in terms of debugging usefulness. Or maybe I'm just not imaginative enough.

asmeurer:
Yeah, uh, I also am not really sure if it would be useful for debugging. Being able to inject arbitrary code into the frame would be useful, but this is not quite that. 

I suppose if you wanted to skip a block of code for some reason, maybe. If I ever find myself wanting this, I'll make a note of it.

flowersw:
Can you also jump back to a specific line in a file you're debugging, like:

> j(ump) lineno
Set the next line that will be executed. Only available in the bottom-most frame. This lets you jump back and execute code again, or jump forward to skip code that you don’t want to run.

It should be noted that not all jumps are allowed – for instance it is not possible to jump into the middle of a for loop or out of a finally clause.

w-hc:
Hi, this would be very useful if the loop is compute-intensive to run, and one only wants to test it for a few before jumping out to test the subsequent code.

asmeurer:
This is straightforward to implement, but I can't figure out how to make the jump command stop on the line it is jumping too, like `t`. It works in `pdb` so I know it should be possible. 

Here is what I have

```diff
diff --git a/pudb/debugger.py b/pudb/debugger.py
index 28a06b5..5f9328e 100644
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -206,6 +206,9 @@ class Debugger(bdb.Bdb):
                     break
                 frame = frame.f_back
 
+    def set_jump(self, frame, line):
+        frame.f_lineno = line
+
     def set_trace(self, frame=None, as_breakpoint=None, paused=True):
         """Start debugging from `frame`.
 
@@ -1161,6 +1164,48 @@ class DebuggerUI(FrameVarInfoKeeper):
[...]

PR #306, opened 2018-07-28
Implement jump command

asmeurer:
~Right now it doesn't stop on the line (like t), which would be ideal. I can't
figure out how to make it do that. pdb's j does it, so it should be possible.~

~It also isn't documented or heavily tested yet.~

This is ready to go.

See issue #129.


asmeurer:
OK, I figured this one out. Turns out `end()` causes it to run past the current line so instead we need to update the UI manually. 

It seems to work now for me. Test script:

```py
def test(n):
    return n + 1

def main():
    for i in range(10):
        if test(i) == 5:
            import pudb;pudb.set_trace()

main()
```

Run the script and jump to the `if test(i) == 5` line. You can then press `s` to step into `test()` and `i` still equals 4.

FGrose:
@asmeurer Thank you! This is helping me greatly to test state and debug.

inducer:
This still has a linter failure. Could you take a look/

asmeurer:
I've fixed the issue.

inducer:
Thanks!

PR #309, opened 2018-08-09
Skip namespace packages in module listing

alexfikl:
Managed to hit this with two modules: `mpl_toolkits` and `sphinxcontrib` that don't have an `__init__.py`. It just started happening after the update to Python 3.7, but I'm not sure what changed.

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -1297,6 +1297,8 @@ class DebuggerUI(FrameVarInfoKeeper):
             def mod_exists(mod):
                 if not hasattr(mod, "__file__"):
                     return False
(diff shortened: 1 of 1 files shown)


inducer:
Thanks for your contribution! For my education, what is the symptom? I imagine you hit Enter on these entries and it dies because there's no file theree?

alexfikl:
The way I triggered it was: 

- `python -m pudb script.py`,
- run to some breakpoint,
- hit `m` to look for a module

which then raised a `TypeError` in `splitext` because it was taking a `None` for those modules.

inducer:
OIC. Thanks for explaining.

Issue #310, opened 2018-09-12
Question: Does pudb support skip?


mapa17:
Does pudb support skipping of modules similar to pdb?

https://docs.python.org/3/library/pdb.html#pdb.Pdb

```python
import pdb; pdb.Pdb(skip=['django.*']).set_trace()
```

thank you,
Manuel

mapa17:
I opened a PR https://github.com/inducer/pudb/pull/311

enabling the option to pass kwargs to the underlying python debugger framework (bdb) through **pudb._get_debugger()**.

The following example shows how keep pudb from entering the decorator library code

```python
import pudb
from configfy import configfy as cfy

# Prevent pudb from stepping into the decorator library code
pudb._get_debugger(skip=['configfy.*'])


@cfy
def fuu(kw_me=42):
    print(kw_me)

if __name__ == '__main__':
    pudb.set_trace()
    fuu()
```

mapa17:
The PR was merged. I take this question to be answered. Thx!

PR #311, opened 2018-09-12
Passing kwargs to python debugger framework

mapa17:
This change enables the passing of kwargs to bdb as discussed in issue #310


inducer:
LGTM. Thanks for your contribution!

Issue #313, opened 2018-09-17
Support debugging from a separate terminal (without using current stdin/stdout).


swarmer:
So using pudb in multithreaded programs is quite painful because secondary threads tend to keep writing shit to stdout and it messes up debugging. What we want then is to access pudb in a separate terminal without involving normal stdin/stdout.

I've experimented a bit with this and thankfully pudb's Debugger has stdin/stdout arguments so this was pretty easy to do. Seems to work ok, except that ipython insists on using the original tty, so I had to pick the internal repl (I haven't looked into this, may be easy to fix).

I want to make my hack more integrated/official/documented. What is involved:  
- Opening a separate terminal in really just a shell one-liner (I used tmux, although this is kind of optional). So It would probably make sense to just describe it in the documentation.
- We need to open tty that we just attached to above and pass it to pudb. I'd add an argument[s?] to `set_trace` and mention this in the doc.

I'll do this if I'll get the go-ahead from maintainers, or if you have any other ideas on how to do this better, I'd love to hear them.

inducer:
I'd be happy to have something like this. If you could somehow automate the process of spawning a tty (or tmux pane) with tmux or another terminal connected in a somewhat convenient manner (`pu.dterm`?), I think this would become a popular feature quite quickly.

swarmer:
I don't think `pu.dterm` can work because we need to create a separate terminal. Opening a new terminal in a cross-platform way sounds like a PITA and even then, which size do we make it? (autoresizing didn't work in my testing, don't know how hard to fix this).

What we can do is: if an env variable, say `PUDB_TTY`, is defined, then `pu.db` instead of using the normal terminal, will use whatever is in this variable. So you would need to open a terminal beforehand and point pudb to that.

Anyway, I'm hoping to make something runnable on this weekend to get feedback.

inducer:
In order to receive `SIGWINCH`, the new terminal would need to become the "controlling terminal" of the process. [Dive in](https://blog.nelhage.com/2011/02/changing-ctty/) if you're curious.

PR #315, opened 2018-09-22
Add ability to control pudb from a separate tty

swarmer:
Alright, here are tweaks that I talked about in #313. The usage is simple: run pudb.create_tty in a separate terminal and before running whatever you're debugging, set the PUDB_TTY environment variable.

I didn't implement autoresizing. I don't want to change the controlling terminal of the program being debugged: what if someone debugs code that is related to just that, or just want to debug signal handlers activated from a terminal, and so on. I'm sure there is a way to make this work at least semi-cleanly, but in any case I can't be bothered to implement this, and it's probably not worth the complexity too.

But on python 3 it does detect the terminal size initially, so you can at least make it as big as you need before entering the debugger.

### DIFF
--- a/doc/starting.rst
+++ b/doc/starting.rst
@@ -52,6 +52,8 @@ for example::
     PUDB_TTY=/dev/pts/3 pudb my-script.py
 
 Now instead of using the current terminal, pudb will use this tty for its UI.
+You may want to use the internal shell in pudb, as others will still use the
+original terminal.
 
 Remote debugging
 ^^^^^^^^^^^^^^^^


pudb/__init__.py
```diff
@@ -70,8 +70,27 @@ def go(self):
...
+    return os.environ.get('PUDB_TTY')
+
+
 def _get_debugger(**kwargs):
     if not CURRENT_DEBUGGER:
+        tty_path = _tty_override()
+        if tty_path:
+            import io, os, sys
+            if sys.version_info[0] == 2:
+                tty_file = open(tty_path, 'r+b')
```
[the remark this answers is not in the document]
inducer: Would it make sense to close/reopen on screen enter/leave? Alternatively, look at `pudb.remote` for inspiration.

inducer:
Could you add some text to the docs (and maybe the README) on how to use this?

asmeurer:
It would be interesting if it could find the tty automatically, perhaps by writing the tty name to the ~/.pudb/ttyname or something. 

It would also be good to have a setting to start tmux using `-CC`. In iTerm2, this enables native tmux mode https://www.iterm2.com/documentation-tmux-integration.html. 

Actually, ideally for me, it would just work completely automatically, so that it just works. In PuDB, there would be a keyboard shortcut or button to open a new terminal and restore the stdout/stdin for the current terminal. Is this possible?

swarmer:
> It would be interesting if it could find the tty automatically, perhaps by writing the tty name to the ~/.pudb/ttyname or something.

I thought about this, but what if you run two debuggers at the same time?

> It would also be good to have a setting to start tmux using `-CC`. In iTerm2, this enables native tmux mode https://www.iterm2.com/documentation-tmux-integration.html.

Can you describe the use case in more detail? I don't see why that would be needed. My original idea is that you create a one-off tmux session just for debugging (it's not even strictly necessary to use tmux, it's just nicer to bail out if you attach from two debuggers at the same time and fuck up you terminal or something). So it's just a terminal in a terminal and it's not supposed to be nice or convenient to manage tabs or anything because you wouldn't create new tabs in this throwaway tmux session.

Now I just thought about another case: if you are ssh'd into a server, then you'd probably want to use a tty in your already existing tmux session. The thing is, again, you don't really have to do any terminal nesting, you can just grab `tty` and use that. Maybe I should document this additionally, or add a flag to my script so it pretty much just does `tty` and sleeps.

> Actually, ideally for me, it would just work completely automatically, so that it just works. In PuDB, there would be a keyboard shortcut or button to open a new terminal and restore the stdout/stdin for the current terminal. Is this possible?

Do you know how to determine which terminal emulator we should use? I'm not aware of a clean way and I think you'll open a big can of worms trying to do that. Even if we test some popular terminals, how do we know on Linux that we need to use $yetanotherterminal by a user's Obscure Desktop Environment no one heard of? Then, what if they have two environments installed, how do we pick between, say, gnome terminal and KDE's one? If they set up one of them to their liking, they will be annoyed if we open the other one (now imaging the stream of bug reports about that). Even on macOS if we use the built-in terminal, someone who uses iTerm and changed hotkeys there will tear their hair out.

asmeurer:
> Can you describe the use case in more detail? I don't see why that would be needed. My original idea is that you create a one-off tmux session just for debugging (it's not even strictly necessary to use tmux, it's just nicer to bail out if you attach from two debuggers at the same time and fuck up you terminal or something). So it's just a terminal in a terminal and it's not supposed to be nice or convenient to manage tabs or anything because you wouldn't create new tabs in this throwaway tmux session.

Even switching between the two is easier with the tmux integration, because it's just like another tab in the GUI. You don't have to learn any tmux commands.

> Do you know how to determine which terminal emulator we should use? 

I think you can detect iTerm2 with an escape sequence. There's probably also an environment variable, but an escape sequence will work even over SSH. But I would just add a setting for it.

swarmer:
> I think you can detect iTerm2 with an escape sequence. There's probably also an environment variable, but an escape sequence will work even over SSH. But I would just add a setting for it.

I don't think we can ever create a terminal on an ssh client machine from inside ssh.

I don't know, selecting from a predefined set of supported terminal emulators in settings sounds alright, but then we'd also need to handle resizing of this separate terminal, otherwise it's a pretty useless feature. Who will implement this? I'm afraid it ain't me, I don't think I can commit enough time to do all of this right now, sorry.

asmeurer:
We may be talking past each other here. I'm not completely sure what sorts of things are possible with different terminal emulators. My only suggestion is to add a flag/setting to add the `-CC` flag to `tmux` (control mode), which iTerm2 automatically detects and does stuff with (possibly other things would too).

swarmer:
Okay, I'll add a flag for -CC and also a flag to use the current terminal (which I wanted for remote servers) when I'll find some time.

doc/starting.rst
```diff
@@ -29,6 +29,24 @@ This is equivalent to::
...
+It's possible to control the debugger from a separate terminal. This is useful
+if there are several threads running that are printing to stdout while
+you're debugging and messing up the terminal, or if you want to keep the
+original terminal available for any other reason.
+
+To create a new terminal, use the utility script (tmux must be installed)::
+
+    python -m pudb.create_tty
+
+Then set the PUDB_TTY environment variable to this path, for example::
```
inducer: What is "this path"?

pudb/create_tty.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+def main():
+    try:
+        subp.call(['tmux', 'new', 'echo -n "Set PUDB_TTY to: "; tty; perl -MPOSIX -e pause'])
+    except OSError as e:
+        if e.errno == errno.ENOENT:
+            print('Error: this script requires tmux to be installed and in PATH')
+        else:
+            raise
```
inducer: I'm not sure this adds much value without an automated way to launch it.

doc/starting.rst
inducer: Maybe simply point out that you can use `tty` to find the device name of an existing pseudoterminal.

pudb/create_tty.py
inducer: (I.e. maybe remove it.)

swarmer:
So @inducer, as I understand you suggest removing the tty creation script and instead documenting that you need to run tty to get the tty name and then not leave an active shell in the terminal so it wouldn't read pudb's input? I think that's also fine.

What remains to be discussed is: are we ok to support the environment variable that I'm using here? I would say yes, and even if someone gets around to implement automated creation of terminals that we discussed here, I still would like to see multiple options to coexist because I personally would still like to setup and keep a single dedicated terminal in which my pudb will open every time right from the start (this is also probably the only sane way to debug on remote servers with many multiplexed panes).

So if I keep the environment variable handling, and improve the docs instead of the new script, does that sound good?

inducer:
> and then not leave an active shell in the terminal so it wouldn't read pudb's input?

Hmm, I had missed that concern. I would still tell prefer documenting what's needed (`perl -MPOSIX -e pause` as you wrote) and letting people create their own shortcuts.

> What remains to be discussed is: are we ok to support the environment variable that I'm using here?

Yeah. I have no objections.

> So if I keep the environment variable handling, and improve the docs instead of the new script, does that sound good?

Yup.

swarmer:
Done

PR #321, opened 2018-10-29
Drop requests from requirements.dev*.txt

inducer:
@asmeurer Do you remember why that was in there in the first place? Github sent me a note about a security vulnerability, which alerted me to its presence.


inducer:
Looks like it was added along with codecov...?

inducer:
Doesn't seem to have any adverse effect. `git grep` comes up empty, too--I say it's gone.

PR #323, opened 2018-11-29
give option to replace tty in set_trace method

idokoren1:


pudb/__init__.py
```diff
@@ -214,15 +218,18 @@ def runcall(*args, **kwds):
...
 
-def set_trace(paused=True):
+def set_trace(paused=True, tty=None):
     """
     Start the debugger
 
     If paused=False (the default is True), the debugger will not stop here
     (same as immediately pressing 'c' to continue).
+
+    tty- Allow the user to control the debugger from seperate terminal given in tty
```
inducer:
```suggestion
:arg tty: Allow the user to control the debugger from seperate terminal given in tty
```

inducer: Also describe how this is given: as an FD, as a filename, ...

inducer:
I'm not sure I'm a fan of this, because it only works the first time around.

idokoren1:
@inducer thanks for the comments.
i rethink about it. 
its possible to control the TTY in run time by using:
os.enviorn["PUDB_TTY"] = "/dev/tty... "

its questionable what is more elegant since the approach in the pull request make this feature more usable (the option for another TTY is quite hidden)
but as long as it possible to control the TTY i'm good with current state.

PR #326, opened 2019-01-25
Add a key to go to the current breakpoint

fdevibe:
When you type 'g' (for go to) in the breakpoints view, display that
breakpoint's file with focus on that breakpoint's line number.


asmeurer:
Enter jumps to the frame in the stack view, so either `g` or `j` would be fine.

inducer:
Thanks for your contribution. I'm sorry I didn't have a ton of bandwidth to review this earlier in the year. Just now, a similar PR showed up (#351) that I like more. Closing here. Again, thanks for your submission.

PR #328, opened 2019-02-12
Add support for scrolling sidebar panes with j/k

craftyguy:
This adds support for scrolling the variable, stack, and breakpoint
panes with the 'j' and 'k' keys. This functionality is disabled by
default, and can be enabled with a new config option in the config
dialog.

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -945,6 +945,9 @@ class DebuggerUI(FrameVarInfoKeeper):
         self.var_list.listen("[", partial(change_rhs_box, 'variables', 0, -1))
         self.var_list.listen("]", partial(change_rhs_box, 'variables', 0, 1))
 
+        self.var_list.listen("j", self.rhs_scroll_down)
+        self.var_list.listen("k", self.rhs_scroll_up)
+
         # }}}
 
         # {{{ stack listeners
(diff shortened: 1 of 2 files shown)


craftyguy:
I could see great possibility in using something similar to this patch for adding/overriding any number of key bindings in the application, so I tried to keep that possibility in mind when writing this.

I'm *very* new to the pudb codebase (first look was last night) , so if there's a better way to implement this and/or some refactoring of this patch is needed, let me know!

craftyguy:
I guess another way to write this patch is to get rid of the SignalWrap methods I added and use rhs_scroll_{up,down} to decide whether or not to send the key up/down events based on the config option.. but this would leave the signal wired up for listening to 'j' and 'k' at all times (even if the scroll option was 'disabled')

inducer:
> but this would leave the signal wired up for listening to 'j' and 'k' at all times (even if the scroll option was 'disabled')

Fine by me (and totally worth it for the simpler logic, iMO).

craftyguy:
@inducer Thanks for the feedback! I have modified this MR to use the simpler behavior, and handle the application of the config in the up/down methods

pudb/debugger.py
```diff
@@ -945,6 +945,9 @@ def insert_watch(w, size, key):
         self.var_list.listen("[", partial(change_rhs_box, 'variables', 0, -1))
         self.var_list.listen("]", partial(change_rhs_box, 'variables', 0, 1))
 
+        self.var_list.toggle("j", partial(self.rhs_scroll_down))
+        self.var_list.toggle("k", partial(self.rhs_scroll_up))
```
inducer: Shouldn't these be `listen`?

craftyguy: Yes, that was my fault. I had apparently tested it before reverting the changes to the SignalWrapper.. Just fixed it!

PR #330, opened 2019-03-11
Only avoid SetPropogatingDict in the builtin shell in Python 2

asmeurer:
The issues with using a custom dict subclass to eval/exec are fixed in
Python 3. Without this, local functions in the shell cannot access global
variables (for instance, inside of list comprehensions).

See the discussion on issues #166 and #103.

To reproduce the issue, debug a function, and make sure you are inside a function frame (not at the top module level). Then type Ctrl-x to go to the shell and enter

```py
x = 5
list(filter(lambda s: x, [{}]))
```

In Python 3, you will see `[{}]`. In Python 2, you get

```
  File "<pudb command line>", line 1, in <module>
  File "<pudb command line>", line 1, in <lambda>
NameError: global name 'x' is not defined
```

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -1576,11 +1576,15 @@ class DebuggerUI(FrameVarInfoKeeper):
             sys.stdin = None
             sys.stderr = sys.stdout = StringIO()
             try:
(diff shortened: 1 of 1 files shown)


inducer:
Thank you very much for working on this!

PR #337, opened 2019-04-24
fix issue with libedit messing up terminal state

### LINKED ISSUE #336: tab completion in internal shell causes escape mode mess
admittedly, in [sabotage linux](https://github.com/sabotage-linux/sabotage) we have a quite unusual python setup:
- [netbsd-curses](https://github.com/sabotage-linux/netbsd-curses) is used instead of ncurses
- libedit is used instead of readline

nonetheless, i'm reporting the issue i experienced with pudb 2018-1 here for the record.

when the internal shell is selected, and one switched there using '!',  as soon as TAB is pressed for the first time for completion (there needs to be something written, for example "a", which would then be completed to e.g. "abs" or "assert"), escapes are broken and every subsequent keypress  is displayed as eg `^CC` or similar, but not recognized as such. makes it impossible to exit the shell via CTRL-X. or use backspace, etc. only way out is to kill pudb.

fortunately, i noticed that if i used e.g. the "classic" shell once, then switched to "internal" using preferences, everything behaves as exactly as it should.
this lead me to produce the following patch, which fixes the issue for me.
apparently the readline completer needs to be "initialized" with the code in the patch at least once.

```diff
--- pudb-2018.1/pudb/debugger.py
+++ pudb-2018.1.org/pudb/debugger.py
@@ -1831,7 +1831,13 @@
 
         def run_cmdline(w, size, key):
             if CONFIG["shell"] == "internal":
-                return toggle_cmdline_focus(w, size, key)
+                curframe = self.debugger.curframe
+                self.screen.stop()
+                import pudb.shell as shell
+                shell.setup_readline(curframe.f_globals, curframe.f_locals)
+                ret = toggle_cmdline_focus(w, size, key)
+                self.screen.start()
+                return ret
             else:
                 return run_external_cmdline(w, size, key)
 
--- pudb-2018.1/pudb/shell.py
+++ pudb-2018.1.org/pudb/shell.py
@@ -72,14 +72,7 @@
 
 custom_shell_dict = {}
 
-
-def run_classic_shell(globals, locals, first_time=[True]):
-    if first_time:
-        banner = "Hit Ctrl-D to return to PuDB."
-        first_time.pop()
-    else:
-        banner = ""
-
+def setup_readline(globals, locals):
     ns = SetPropagatingDict([locals, globals], locals)
 
     from pudb.settings import get_save_config_path
@@ -97,6 +90,18 @@
             readline.read_history_file(hist_file)
         except IOError:
             pass
+
+    return hist_file
+
+
+def run_classic_shell(globals, locals, first_time=[True]):
+    if first_time:
+        banner = "Hit Ctrl-D to return to PuDB."
+        first_time.pop()
+    else:
+        banner = ""
+
+    hist_file = setup_readline(globals, locals)
 
     from code import InteractiveConsole
     cons = InteractiveConsole(ns)
```
(the self.screen.stop/start is probably not necessary, it's a leftover from my first attempts at fixing)

rofl0r:
when switching to the internal shell and using tab completion, libedit messes
up the terminal state, unless it uses the readline initialization routine that
the "classic" shell code uses as well.
additionally to that, it requires that the console screen is stopped before
doing that for the first time.

fixes #336

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -1853,9 +1853,21 @@ class DebuggerUI(FrameVarInfoKeeper):
 
             self.update_var_view()
 
+        def run_internal_shell(w, size, key, first_time=[True]):
+            frame = self.debugger.curframe
+            if first_time: self.screen.stop()
+            import pudb.shell as sh
+            histfile, ns = sh.readline_init(frame.f_globals, frame.f_locals)
+            ret = toggle_cmdline_focus(w, size, key)
+            sh.readline_fini(histfile)
+            if first_time:
+                self.screen.start()
+                first_time.pop()
+            return ret
+
         def run_cmdline(w, size, key):
             if CONFIG["shell"] == "internal":
-                return toggle_cmdline_focus(w, size, key)
+                return run_internal_shell(w, size, key)
             else:
                 return run_external_cmdline(w, size, key)
 
--- a/pudb/shell.py
+++ b/pudb/shell.py
@@ -72,14 +72,7 @@ class SetPropagatingDict(dict):
 
 custom_shell_dict = {}
 
-
-def run_classic_shell(globals, locals, first_time=[True]):
-    if first_time:
-        banner = "Hit Ctrl-D to return to PuDB."
-        first_time.pop()
-    else:
-        banner = ""
-
+def readline_init(globals, locals):
     ns = SetPropagatingDict([locals, globals], locals)
 
     from pudb.settings import get_save_config_path
@@ -98,14 +91,27 @@ def run_classic_shell(globals, locals, first_time=[True]):
         except IOError:
             pass
 
+    return hist_file, ns
+
+def readline_fini(hist_file):
+    if HAVE_READLINE:
+        readline.write_history_file(hist_file)
+
+def run_classic_shell(globals, locals, first_time=[True]):
+    if first_time:
+        banner = "Hit Ctrl-D to return to PuDB."
+        first_time.pop()
+    else:
+        banner = ""
+
+    hist_file, ns = readline_init(globals, locals)
+
     from code import InteractiveConsole
     cons = InteractiveConsole(ns)
 
     cons.interact(banner)
 
-    if HAVE_READLINE:
-        readline.write_history_file(hist_file)
-
+    readline_fini(hist_file)
 
 def run_bpython_shell(globals, locals):
     ns = SetPropagatingDict([locals, globals], locals)


asmeurer:
It seems the issue at https://github.com/inducer/pudb/issues/166 is also caused by libedit. I'm curious if you see that at all. I checked and this PR doesn't fix it. 

To reproduce, type 

```
Ctrl-x
1<ENTER>
Ctrl-x
```

If the issue occurs, it will enter `^X` in the shell instead of leaving it, and all other keyboard input will be broken. 

IMHO we should not use readline at all in the internal shell (it isn't used for keyboard input, is it?). Instead we should just implement basic completion natively. I guess we would also lose history support, although I'm not really sure about that because for me it doesn't persist any shell history between sessions. So maybe it's broken anyway (?)

OTOH actually using it for input would be nice. That would enable more readline bindings that don't work, like basic emacs-style word movement.

pudb/debugger.py
```diff
@@ -1853,9 +1853,21 @@ def run_external_cmdline(w, size, key):
 
             self.update_var_view()
 
+        def run_internal_shell(w, size, key, first_time=[True]):
+            frame = self.debugger.curframe
+            if first_time: self.screen.stop()
+            import pudb.shell as sh
+            histfile, ns = sh.readline_init(frame.f_globals, frame.f_locals)
+            ret = toggle_cmdline_focus(w, size, key)
+            sh.readline_fini(histfile)
```
inducer: Do we need to call `readline_init`/`readline_fini` here at all?

inducer:
> OTOH actually using it for input would be nice.

Not very easy to integrate with Urwid though. Not impossible, but potentially a considerable headache.

> IMHO we should not use readline at all in the internal shell (it isn't used for keyboard input, is it?).

The only connection between readline and the internal shell is that the internal shell uses the `rlcompleter` module, which imports readline.

asmeurer:
Apparently just importing `readline` at all causes the problem. I removed the rlcompleter stuff and it still happened just from importing `readline`. I don't know if we can detect if readline is `libedit` without actually importing it. Although I tried importing `readline` in the debugged script and it didn't reproduce the issue.

Also `rlcompleter` imports `readline` [at import time](https://github.com/python/cpython/blob/6aeb2fe606408aae14c246470794f1303b3be812/Lib/rlcompleter.py#L200). So if we want to use its completer, we would need to copy the Completer code from rlcompleter to pudb. 

> The only connection between readline and the internal shell is that the internal shell uses the rlcompleter module, which imports readline.

We supposedly use readline for shell history reading and writing too, but as I noted, it doesn't seem to work.

inducer:
> we would need to copy the Completer code from rlcompleter to pudb.

Yuck. :/ I guess that's still a bit cleaner (as long as we keep the code unmodified) than some `sys.modules` hackery.

> We supposedly use readline for shell history reading and writing too, but as I noted, it doesn't seem to work.

The built-in shell has a separate history mechanism though. Not sure it should be that way, but that's the intention at least. I don't think that mechanism persists history ATM, that should probably be fixed.

asmeurer:
Oh I was confusing `run_classic_shell` with the builtin shell. The explains a lot. I guess it was confusing me because those imports in shell.py are what affect this. I actually didn't touch the `rlcompleter` import in debugger.py at all. So maybe the issue only occurs if readline isn't imported right away (at the module level).

asmeurer:
Yeah, this fixes #166 (no idea about #336)

```diff
diff --git a/pudb/shell.py b/pudb/shell.py
index 4d5763c..8917728 100644
--- a/pudb/shell.py
+++ b/pudb/shell.py
@@ -26,14 +26,6 @@ else:
     HAVE_PTPYTHON = True


-try:
-    import readline
-    import rlcompleter
-    HAVE_READLINE = True
-except ImportError:
-    HAVE_READLINE = False
-
-
 # {{{ combined locals/globals dict

 class SetPropagatingDict(dict):
@@ -88,6 +80,13 @@ def run_classic_shell(globals, locals, first_time=[True]):
             get_save_config_path(),
             "shell-history")

+    try:
+        import readline
+        import rlcompleter
+        HAVE_READLINE = True
+    except ImportError:
+        HAVE_READLINE = False
+
     if HAVE_READLINE:
         readline.set_completer(
                 rlcompleter.Completer(ns).complete)
```

asmeurer:
I wonder if we can actually use the SetPropogatingDict in Python 2 https://github.com/inducer/pudb/pull/330. Supposedly it doesn't work to pass a dict subclass to exec in Python 2, so maybe not.

rofl0r:
>Also rlcompleter imports readline at import time. So if we want to use its completer, we would need to copy the Completer code from rlcompleter to pudb.

one should think there's a pure python module that handles autocompletion/history without resorting to terminal escape hackery...

PR #344, opened 2019-05-30
Detect external tty size on python 2

flyaroundme:
I've recently discovered an issue: the external terminal size is pretty small when running pudb against python 2.
Here's the quick fix for it based on #315 and the idea from #209


inducer:
Thanks!

PR #350, opened 2019-07-09
Move the readline import out of the top (module) level and add Jedi completion

### LINKED ISSUE #166: Ctrl-x not function to leave console

asmeurer:
This fixes #166. I have no idea why, but if readline is really libedit (like
for a system or python.org install on Mac), it causes the built in shell to
break the terminal inputs, but only when imported at the top level. If the
import is deferred, as far as I've seen, it doesn't happen.

CC @rofl0r

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -1492,7 +1492,11 @@ class DebuggerUI(FrameVarInfoKeeper):
                     coming_from="above")
 
         def cmdline_tab_complete(w, size, key):
-            from rlcompleter import Completer
+            try:
+                from jedi import Interpreter
+            except ImportError:
+                add_cmdline_content("Tab completion requires jedi to be installed.",
+                                    "command line output")
 
             text = self.cmdline_edit.edit_text
             pos = self.cmdline_edit.edit_pos
@@ -1500,30 +1504,9 @@ class DebuggerUI(FrameVarInfoKeeper):
             chopped_text = text[:pos]
             suffix = text[pos:]
 
-            # stolen from readline in the Python interactive shell
-            delimiters = " \t\n`~!@#$%^&*()-=+[{]}\\|;:\'\",<>/?"
-
-            complete_start_index = max(
-                    chopped_text.rfind(delim_i)
-                    for delim_i in delimiters)
-
-            if complete_start_index == -1:
-                prefix = ""
-            else:
-                prefix = chopped_text[:complete_start_index+1]
-                chopped_text = chopped_text[complete_start_index+1:]
-
-            state = 0
-            chopped_completions = []
-            completer = Completer(cmdline_get_namespace())
-            while True:
-                completion = completer.complete(chopped_text, state)
-
-                if not isinstance(completion, str):
-                    break
-
-                chopped_completions.append(completion)
-                state += 1
+            completions = Interpreter(chopped_text, [cmdline_get_namespace()]).completions()
+            full_completions = [i.name_with_symbols for i in completions]
+            chopped_completions = [i.complete for i in completions]
 
             def common_prefix(a, b):
                 for i, (a_i, b_i) in enumerate(zip(a, b)):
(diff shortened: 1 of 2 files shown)


asmeurer:
I wonder if we can actually use the SetPropogatingDict in Python 2 #330. Supposedly it doesn't work to pass a dict subclass to exec in Python 2, so maybe not.

inducer:
Thanks for this PR. I'm mostly OK with this, but I have two comments:

* I'd like it if a comment were added to the import with a link to this PR.
* Wouldn't this still import `readline` from [here](https://github.com/inducer/pudb/blob/a4f80c1e9ca6adb7489676ff5b531ef603799afa/pudb/debugger.py#L1495) via `rlcompleter`? (Once Tab is pressed?) Or is that harmless?

asmeurer:
Ah you're right. If you `C-x 1 TAB` with this PR is still messes up.

asmeurer:
I've pushed a commit that uses jedi to do completion, if it is installed. We can also extract the rlcompleter completer and use it as a fallback if you like. I would try to do it without importing readline if possible, to avoid this stupid libedit issue. Unless someone knows of a way to detect libedit without importing readline (or if someone can figure out why importing libedit readline breaks urwid and how to fix it).

asmeurer:
We can copy this if we want basic `dir` completion https://github.com/asmeurer/mypython/blob/master/mypython/dircompletion.py.

pudb/debugger.py
```diff
@@ -1492,38 +1492,25 @@ def add_cmdline_content(s, attr):
...
-                completion = completer.complete(chopped_text, state)
-
-                if not isinstance(completion, str):
-                    break
-
-                chopped_completions.append(completion)
-                state += 1
+            try:
+                completions = Interpreter(chopped_text, [cmdline_get_namespace()]).completions()
+            except:
```
inducer: Flake8 is going to want `except Exception:`

pudb/debugger.py
```diff
@@ -1492,38 +1492,25 @@ def add_cmdline_content(s, attr):
...
+                # Jedi sometimes produces errors. Ignore
```
inducer: How frequent are those? Could we at least add an advisory error message to the console output if that happens? Because, you know, "errors should never pass silently"? :)

pudb/debugger.py
```diff
@@ -1492,38 +1492,25 @@ def add_cmdline_content(s, attr):
...
-            chopped_completions = []
-            completer = Completer(cmdline_get_namespace())
-            while True:
-                completion = completer.complete(chopped_text, state)
-
-                if not isinstance(completion, str):
-                    break
-
-                chopped_completions.append(completion)
-                state += 1
```
inducer: Does Jedi not need all this logic?

asmeurer: No jedi just returns a list of completions.

pudb/debugger.py
asmeurer: I don't know what it's like these days. I can try printing them and we can see.

asmeurer: I made it print the error. I don't know if jedi is actually better these days, so we'll see if it actually happens a lot or not. I also don't know to what degree it actually executes stuff like `dir()` on the objects in the namespace. The rlcompleter definitely does that, which can cause arbitrary code execution if the objects do certain things (like define `__dir__`).

inducer:
This looks good to me. Ready to merge when you are.

asmeurer:
Let's see if the person who reported https://github.com/inducer/pudb/issues/352 or one of the other issues is able to test this first.

inducer:
Based on all the feedback we've received, it seems like this might be a win. @asmeurer, thoughts?

asmeurer:
Yeah I think we've had enough people confirm that this fixes this.

PR #351, opened 2019-07-12
Improve help and add jump to breakpoint

bulletmark:
Please see the words on the 2 commits here.

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -44,6 +44,11 @@ from pudb.py3compat import PY3, raw_input, execfile
 CONFIG = load_config()
 save_config(CONFIG)
 
+HELP_HEADER = r"""
+Key Assignments: Use Arrow Down/Up or Page Down/Up to scroll.
+
+"""
+
 HELP_MAIN = r"""
 Keys:
     Ctrl-p - edit preferences
@@ -932,7 +937,7 @@ class DebuggerUI(FrameVarInfoKeeper):
                 self.update_var_view()
 
         def helpside(w, size, key):
-            help(HELP_SIDE + HELP_MAIN + HELP_LICENSE)
+            help(HELP_HEADER + HELP_SIDE + HELP_MAIN + HELP_LICENSE)
 
         self.var_list.listen("\\", change_var_state)
         self.var_list.listen(" ", change_var_state)
@@ -1449,7 +1454,7 @@ class DebuggerUI(FrameVarInfoKeeper):
                                 self.translate_ui_stack_index(pos))
 
         def helpmain(w, size, key):
-            help(HELP_MAIN + HELP_SIDE + HELP_LICENSE)
+            help(HELP_HEADER + HELP_MAIN + HELP_SIDE + HELP_LICENSE)
 
         self.source_sigwrap.listen("n", next)
         self.source_sigwrap.listen("s", step)
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -101,7 +101,7 @@ Sidebar-related (active in sidebar):
     [/] - grow/shrink relative size of active sidebar box
 
 Keys in variables list:
-    \/space - expand/collapse
+    \/enter/space - expand/collapse
     t/r/s/c - show type/repr/str/custom for this variable
     h - toggle highlighting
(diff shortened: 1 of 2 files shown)


bulletmark:
After issuing this PR here I have just noticed PR #326 which implements similar code for the jump function. We both essentially copied some of the existing code elsewhere in that module to do the jump but note I think my implementation is better for 2 reasons:

He included the `set_focus()` line but we don't really want that because we want the BP list to work the same as the stack list, i.e. you can scroll up/down and just press enter to quickly see where the BP/stack is without having to navigate back to the BP/stack list window. For the same reason I use "enter" to do this rather than a specific key because that is how the stack list works so it is more consistent and more intuitive. I remapped "e" to do the old "edit" function and made "b" to toggle the BP state because that is the key users are familiar with from the main code window , so again that is more consistent anyhow.

pudb/debugger.py
```diff
@@ -44,12 +44,7 @@
 CONFIG = load_config()
 save_config(CONFIG)
 
-HELP_TEXT = r"""\
-Welcome to PuDB, the Python Urwid debugger.
--------------------------------------------
-
-(This help screen is scrollable. Hit Page Down to see more.)
```
inducer: How come you consider this line expendable?

inducer:
I like your approach better than #326, so I'd say let's focus here. My main quibble is the removal of the introductory help text.

pudb/debugger.py
bulletmark: The first line I didn't really delete because I moved it to the title bar. The "is scrollable" line I consider redundant. It's help text, users automatically know it will be scrollable. I also deleted a redundant "\" in the first line which appears to be an error.

I am a new user and new users just want to pop up the help page quickly to see the key assignments and no other cruft, without having to scroll around. I don't think valuable screen real estate should be wasted.

I personally don't think the license should even be in the help page at all but as a concession to your existing functionality I left it at the bottom on both help pages.

bulletmark:
Note in my personal branch (not pushed here) I also change the Variables list so that "enter" just expands like "/" or " " because, like the Stack and Breakpoint lists, I want "enter" to do the most intuitive and most common function users want to do. I have not pushed that change here because I imagine that you and existing users may want to keep "enter" for the edit function. As a new user, I found I was constantly pressing "enter" in the Variables list and then having to close down that popup in frustration every time, before I remembered I am supposed to press " ". Expanding a variable is by far the most common function people do in that list so I think that should be the "enter" function.

pudb/debugger.py
asmeurer: I think the scrolling instructions were put there because it was not clear to some people that this was possible. Unfortunately urwid lacks in discoverability sometimes.

inducer:
> Note in my personal branch (not pushed here) I also change the Variables list so that "enter" just expands like "/" or " " because, like the Stack and Breakpoint lists, I want "enter" to do the most intuitive and most common function users want to do.

Sure, I can see that argument. Feel free to include that here.

pudb/debugger.py
```diff
@@ -780,7 +780,7 @@ def change_var_state(w, size, key):
             iinfo = self.get_frame_var_info(read_only=False) \
                     .get_inspect_info(var.id_path, read_only=False)
 
-            if key == "\\" or key == ' ':
+            if key == "enter" or key == "\\" or key == ' ':
```
inducer: Could you update the help text to reflect these changes?

bulletmark: Sorry about that. Done now.

pudb/debugger.py
inducer: To be clearer: Could you please add this back?

bulletmark: As I said, "PuDB - The Python Urwid Debugger" is in the title now so surely you don't want that? Also, surely you don't want to waste real-estate on that '------' line? So are you only wanting the "is scrollable" line?

bulletmark: Does commit 7aeadbe159387b6403ced25d57d41331b71483b6 suffice?

inducer: Sure, that's fine.

inducer:
Thanks for your contribution!

PR #353, opened 2019-07-15
Add option to hide command window when not in use

bulletmark:
I and I suspect many other users don't use the command line window at the bottom so this PR adds an option to hide it by default. It pops up in to view when you Ctrl-X to it, and hides when you Ctrl-X back to the source window. Apart from the room to view your code, I think pudb looks aesthetically better and more balanced with a single source window on the left, and the sidebar windows on the right. With the new option set off, pudb displays the command window all the time as it does now.

Also removed the redundant "jk_sidebar_scroll" option. jk scroll should just be always enabled in sidebars like in main source window. Happy to put this back if somebody can explain the point of needing it?

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -700,10 +700,12 @@ class DebuggerUI(FrameVarInfoKeeper):
         self.cmdline_sigwrap = SignalWrap(
                 urwid.AttrMap(self.cmdline_pile, None, "focused sidebar")
                 )
-
+        self.cmdline_on = not CONFIG["hide_cmdline_win"]
+        self.cmdline_weight = 1
         self.lhs_col = urwid.Pile([
             ("weight", 5, self.source_attr),
-            ("weight", 1, self.cmdline_sigwrap),
+            ("weight", self.cmdline_weight if self.cmdline_on else 0,
+                self.cmdline_sigwrap),
             ])
 
         # }}}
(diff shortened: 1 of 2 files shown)


bulletmark:
Note I made the default (for new users) to hide this window although I can anticipate you may want to change this to un-hidden (i.e. a one word code change). I just think new users don't want to bother with that window so it makes pudb look cleaner on very first use.

bulletmark:
I should add that this PR respects and remembers the window size. So if you Ctrl-X open it, then resize it larger, the Ctrl-X close it then it will still be the larger size when you Ctrl-X back to it.

inducer:
Just played with this. It feels nice, I like it, and I have only one comment about the code: To maintain discoverability of the internal shell, I think I'd like to have the auto-hide feature default to off. @asmeurer, how do you feel about this?

asmeurer:
Yeah I agree it should be shown by default. Otherwise people won't know it is there. I have no idea about the jk stuff.

asmeurer:
I was going to suggest putting "Ctrl-x" at the top with the other shortcuts but it's already full for a default 80 character wide terminal, which I think many people use.

inducer:
After deleting the option from my config file, PuDB still starts up with the internal shell window hidden for me. Could you take a look?

bulletmark:
> After deleting the option from my config file, PuDB still starts up with the internal shell window hidden for me. Could you take a look?

Works fine for me. I delete the option (or I completely delete the config file) and the window is shown. Can you please double check you are deleting "hide_command_win".

inducer:
Odd. I must have seen a ghost. Works for me now. Thanks again for your contribution!

inducer:
(For posterity: I did see and think about the removal of the j/k option stuff, I looked briefly at when it was introduced, and I didn't see a compelling reason for keeping it.)

Issue #92, opened 2013-11-05
Improve breakpoints view


inducer:
@asmeurer said in #90: (since moved from there to here)

> I actually realized that the breakpoints view could use some work anyway. I've never been a big user of breakpoints, though so I haven't noticed these before. There should be a visual indication on the breakpoint if it is active or not (I'm thinking some character in the already empty space to the left of each breakpoint). A keyboard shortcut to disable a breakpoint without deleting it would be useful, and a keyboard shortcut to jump to the location might be nice too. Any thoughts on this?  
> 
> I'm thinking:
> - `*` (to the left of a breakpoint) means it is active
> - `s` (to the left of the breakpoint, next to the `*`) means it is a `set_trace` breakpoint. We could put other letters there too to indicate state about the breakpoint if it's useful.
> - `b` (?) disables/enables the breakpoint. Maybe another key is better, but this matches the key used in the code view.
> - `l` jumps to the location of the breakpoint (also not sure about this, so feel free to suggest something better. Maybe `L` is better since that matches the "goto" function in the code).

fdevibe:
OK, only saw this one after I added PR #326, so maybe this is unwanted. However, I didn't see how to achieve this until I read this issue, and I would really like a direct route to the file:line in question. I don't care much about which character is used, though.

mvanderkamp:
This can be closed, no? Only feature not present is the `set_trace` distinction which I don't think is important or really all that useful.

inducer:
I'd say so. Thanks for your work towards this! If anyone disagrees, please speak up.

PR #361, opened 2019-09-17
Add keyboard shortcut l in breakpoint list to show the location

asmeurer:
c.f. #92

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -1109,6 +1109,17 @@ class DebuggerUI(FrameVarInfoKeeper):
                 else:
                     self.update_breakpoints()
 
+        def breakpoint_location(w, size, key):
+            bp_entry, pos = self.bp_list._w.get_focus()
+
+            if bp_entry is None:
+                return
+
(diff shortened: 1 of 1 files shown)


asmeurer:
Not sure if it should be `L` instead, to match the "goto line" shortcut. Personally I always expect `Enter` to do this, but we probably shouldn't change what that already does.

pudb/debugger.py
```diff
@@ -1109,6 +1110,17 @@ def examine_breakpoint(w, size, key):
                 else:
                     self.update_breakpoints()
 
+        def breakpoint_location(w, size, key):
```
inducer: I think this is largely identical to `show_breakpoint`, bound to `Enter` in that context.

asmeurer: Hmm, am I crazy? I was pressing Enter before, but it opened the dialog. I just tested it now and it does what I want. We definitely don't need this PR if Enter already does this.

asmeurer: Oh this was changed recently and I was using the released version. Sorry for the noise. This can be closed.

PR #362, opened 2019-10-10
Fix errant command line expose after search

bulletmark:
In PR #353 I added the option to hide the command line window but I just
noticed that the window errantly appeared if you did a search. This
commit fixes that by changing the logic to actually create/delete the
window (like all other PUDB "hidden" windows) instead of merely setting
the display width to 0 as I was doing previously. It seems this is necessary
because URWID sometimes rebiases the widths itself and thus a 0 width
may get set to a non 0 value and thus be exposed.

I also fixed a bug which has always existed, even prior to my first PR
and that is if you opened a search, then did a ctrl+x to the command
window and then again back to the source window then the source window
would get focus instead of the search window. This commit fixes it so
that focus goes back to the search window, if it is still open.

The change to ui_tools.py is merely to remove a redundant "else" that I
noticed.

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -701,13 +701,12 @@ class DebuggerUI(FrameVarInfoKeeper):
         self.cmdline_sigwrap = SignalWrap(
                 urwid.AttrMap(self.cmdline_pile, None, "focused sidebar")
                 )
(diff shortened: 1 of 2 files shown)


bulletmark:
Actually, I have changed my mind about how best to address this so closing this PR and will raise another one.

PR #363, opened 2019-10-11
Fix errant command line expose after search

bulletmark:
In PR #353 I added the option to hide the command line window but I recently noticed that the window errantly appeared if you did a search. This commit fixes that by resetting the command line window weight to 0 if it should be hidden, after a search is activated. It seems this is necessary because URWID sometimes rebiases the widths itself and thus a 0 width may get set to a non 0 value and would be exposed.

I also fixed a bug which has always existed, even prior to my first PR and that is if you opened a search, then did a ctrl+x to the command window and then again back to the source window then the source window would get focus instead of the search window and then the search gets locked to an odd state. This commit fixes it so that focus goes back to the search window, if it is still open.

There is also a small change to ui_tools.py is merely to remove a redundant "else" that I noticed and some misc other tiny aesthetic code improvements that I noticed in debugger.py.

This PR completely supercedes my previous PR #362  fix for these issues because PR #362 did not correctly maintain the size of the command window if it was re-opened.

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -762,7 +762,7 @@ class DebuggerUI(FrameVarInfoKeeper):
         def change_rhs_box(name, index, direction, w, size, key):
             from pudb.settings import save_config
 
-            _, weight = self.rhs_col.item_types[index]
+            weight = self.rhs_col.item_types[index][1]
 
             if direction < 0:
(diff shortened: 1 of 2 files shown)


inducer:
LGTM, thanks for your contribution!

PR #365, opened 2019-11-13
Show the value of numpy scalars when using the safe stringifier

asmeurer:
A numpy scalar value is already included in the tests (the test doesn't test the exact output, only that it doesn't crash).


inducer:
Thanks!

PR #367, opened 2019-11-20
Fix the logic for getting line numbers in Python 3.8

### LINKED ISSUE #366: Impossible to set breakpoint in Python 3.8
Python 3.8 + PuDB 2019.1  (but it works in python 3.6, and with pdb python 3.8)

When I try to set breakpoint I receive modal form "Possibly Invalid Breakpoint" (also "run to cursor" doesn't work).
So it's impossible set breakpoint after slash `\` (after 3rd line)
Source:
```python
def main():
    import pudb; pu.db
    conf = ''. \
        replace('', '')

    conf_tpl = ''  # <-- imposible to set breakpoint here

main()
```

asmeurer:
This fixes an issue in Python 3.8 where some lines couldn't be breakpointed.

Fixes #366.

I tested this on the example file from the issue in 3.8, 3.7, and 3.5. I didn't do any other tests, so it's possible this breaks something. We also maybe should add a test for this to the test suite.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -9,6 +9,8 @@ jobs:
                 python.version: '2.7'
             Python36:
                 python.version: '3.6'
+            Python37:
+                python.version: '3.7'
 
     steps:
     -
@@ -38,8 +40,8 @@ jobs:
         vmImage: 'ubuntu-latest'
     strategy:
         matrix:
-            Python37:
-                python.version: '3.7'
+            Python38:
+                python.version: '3.8'
 
     steps:
     -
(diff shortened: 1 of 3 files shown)


inducer:
Thanks for the patch! I've simplified the logic a bit, done some version archaeology and added a test.

inducer:
Given that this is a 3.8-related problem, it makes sense to add 3.8 CI jobs. I've done that, too. I should also probably release a new version, so that not too many folks on 3.8 get caught by this.

inducer:
2019.2, out now.

Issue #382, opened 2020-04-17
set_interrupt_handler documentation


dansebcar:
# Background

I'm trying to use pudb with a web server behind gunicorn, which sets up lots of signal handlers (including SIGINT), and I want pudb to use a different one.

# Problem

In the docstring for `set_interrupt_handler` I'm advised to set `pudb.DEFAULT_SIGNAL` to achieve this. That doesn't currently work, because when I import pudb, `pudb.set_interrupt_handler.__defaults__` contains a reference to SIGINT, not pudb.DEFAULT_SIGNAL (`set_interrupt_handler`'s default arguments are not re-evaluated when it is called, they are only evaluated at import time).

# Proposed solution

Assigning `pudb.set_interrupt_handler.__defaults__ = (signal.SIGUSR1,)` works well for me with the current code, would it be possible to update the docstring to reflect that?

It looks like set_interrupt_handler is never called with any arguments, so an alternative solution would be to enable the documented API would be to remove the parameter and always refer to DEFAULT_SIGNAL directly.

inducer:
#382

inducer:
#383

PR #383, opened 2020-04-19
Fix overridability of default signal in set_interrupt handler (Closes…

inducer:
…: gh-382)


inducer:
@asmeurer Does this look OK to you?

asmeurer:
Ah, that's the problem with APIs that have you override a module's globals. This should fix the issue.

Issue #389, opened 2020-06-15
Support running as module (`pudb3 -m some_module`)


eliasdorneles:
It would be nice if we could run `pudb3 -m some_module ....` similarly to how we can do with pdb (`python -m pdb -m some_module`).

This is useful when you want to debug some code not necessarily written by you, which you must run with `python -m module`.

inducer:
I agree. I'd be happy to take a patch.

eliasdorneles:
Closing this, now that #390 is merged :)

PR #390, opened 2020-06-15
Add -m switch to run as module

eliasdorneles:
Hello,

Here is an implementation for the `-m module` switch (see https://github.com/inducer/pudb/issues/389).

Does this look good?

### DIFF
--- a/pudb/__init__.py
+++ b/pudb/__init__.py
@@ -120,7 +120,13 @@ DEFAULT_SIGNAL = signal.SIGINT
 del signal
 
 
-def runscript(mainpyfile, args=None, pre_run="", steal_output=False):
+def runmodule(*args, **kwargs):
+    kwargs['run_as_module'] = True
+    runscript(*args, **kwargs)
+
+
+def runscript(mainpyfile, args=None, pre_run="", steal_output=False,
+              run_as_module=False):
     dbg = _get_debugger(steal_output=steal_output)
 
     # Note on saving/restoring sys.argv: it's a good idea when sys.argv was
@@ -132,7 +138,10 @@ def runscript(mainpyfile, args=None, pre_run="", steal_output=False):
     import sys
     if args is not None:
         prev_sys_argv = sys.argv[:]
-        sys.argv = [mainpyfile] + args
+        if run_as_module:
+            sys.argv = args
+        else:
+            sys.argv = [mainpyfile] + args
 
     # replace pudb's dir with script's dir in front of module search path.
     from os.path import dirname
@@ -150,11 +159,19 @@ def runscript(mainpyfile, args=None, pre_run="", steal_output=False):
         status_msg = ""
 
         try:
-            dbg._runscript(mainpyfile)
-        except SystemExit:
-            se = sys.exc_info()[1]
-            status_msg = "The debuggee exited normally with " \
-                    "status code %s.\n\n" % se.code
+            if run_as_module:
+                try:
+                    dbg._runmodule(mainpyfile)
+                except ImportError as e:
+                    print(e, file=sys.stderr)
+                    sys.exit(1)
+            else:
+                try:
+                    dbg._runscript(mainpyfile)
+                except SystemExit:
+                    se = sys.exc_info()[1]
+                    status_msg = "The debuggee exited normally with " \
+                            "status code %s.\n\n" % se.code
         except Exception:
             dbg.post_mortem = True
             dbg.interaction(None, sys.exc_info())
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -486,6 +486,37 @@ class Debugger(bdb.Bdb):
         # Implicitly runs in the namespace of __main__.
         self.run(statement)
 
+    def _runmodule(self, module_name):
+        # This is basically stolen from the pdb._runmodule from CPython 3.8
+        # and adapted to work also in Python 2
+        # https://github.com/python/cpython/blob/a1d3be4623c8ec7069bd34ccdce336be9cdeb644/Lib/pdb.py#L1530
+        import runpy
+
+        # here we unpack the module details manually, so that it works in PY2 as well
+        mod_details = runpy._get_module_details(module_name)
+        mod_spec = mod_details[1]
+        code = mod_details[2]
+
+        self.mainpyfile = self.canonic(code.co_filename)
+        import __main__
+        __main__.__dict__.clear()
+        __main__.__dict__.update({
+            "__name__": "__main__",
+            "__file__": self.mainpyfile,
+            "__spec__": mod_spec,
+            "__builtins__": __builtins__,
+        })
+
+        if PY3:
+            __main__.__dict__.update({
+                "__package__": mod_spec.parent,
+                "__loader__": mod_spec.loader,
+            })
+
+        self._wait_for_mainpyfile = True
+
+        self.run(code)
+
 # }}}
 
 
--- a/pudb/run.py
+++ b/pudb/run.py
@@ -4,41 +4,48 @@ from __future__ import absolute_import, division, print_function
 def main():
     import sys
 
-    from optparse import OptionParser
-    parser = OptionParser(
-            usage="usage: %prog [options] [SCRIPT-TO-RUN|-m MODULE] [ARGUMENTS]")
+    import argparse
+    parser = argparse.ArgumentParser(
+        usage="%(prog)s [options] [-m] SCRIPT-OR-MODULE-TO-RUN [SCRIPT_ARGS]"
+    )
+    parser.add_argument("-s", "--steal-output", action="store_true"),
 
-    parser.add_option("-s", "--steal-output", action="store_true"),
-    parser.add_option("-m", "--module", metavar="MODULE",
-                      help="Debug module or package instead of script"),
-    parser.add_option("--pre-run", metavar="COMMAND",
+    # note: we're implementing -m as a boolean flag, mimicking pdb's behavior,
+    # and makes it possible without much fuss to support cases like:
+    #    python -m pudb -m http.server -h
+    # where the -h will be passed to the http.server module
+    parser.add_argument("-m", "--module", action='store_true',
+                      help="Debug as module or package instead of as a script")
+
+    parser.add_argument("--pre-run", metavar="COMMAND",
             help="Run command before each program run",
             default="")
-    parser.disable_interspersed_args()
-    options, args = parser.parse_args()
+    parser.add_argument('script_args', nargs=argparse.REMAINDER,
+                        help="Arguments to pass to script or module")
+
+    options = parser.parse_args()
+    args = options.script_args
 
     options_kwargs = {
         'pre_run': options.pre_run,
         'steal_output': options.steal_output,
     }
 
+    if len(args) < 1:
+        parser.print_help()
+        sys.exit(2)
+
+    mainpyfile = args[0]
+    sys.argv = args
+
     if options.module:
-        sys.argv = args
         from pudb import runmodule
-        runmodule(options.module, **options_kwargs)
+        runmodule(mainpyfile, **options_kwargs)
     else:
-        if len(args) < 1:
-            parser.print_help()
-            sys.exit(2)
-
-        mainpyfile = args[0]
-
         from os.path import exists
         if not exists(mainpyfile):
-            print('Error: %s does not exist' % mainpyfile)
+            print('Error: %s does not exist' % mainpyfile, file=sys.stderr)
             sys.exit(1)
-
-        sys.argv = args
 
         from pudb import runscript
         runscript(mainpyfile, **options_kwargs)
--- a/pudb/run.py
+++ b/pudb/run.py
@@ -6,32 +6,42 @@ def main():
 
     from optparse import OptionParser
     parser = OptionParser(
-            usage="usage: %prog [options] SCRIPT-TO-RUN [SCRIPT-ARGUMENTS]")
+            usage="usage: %prog [options] [SCRIPT-TO-RUN|-m MODULE] [ARGUMENTS]")
 
     parser.add_option("-s", "--steal-output", action="store_true"),
+    parser.add_option("-m", "--module", metavar="MODULE",
+                      help="Debug module or package instead of script"),
     parser.add_option("--pre-run", metavar="COMMAND",
             help="Run command before each program run",
             default="")
     parser.disable_interspersed_args()
     options, args = parser.parse_args()
 
-    if len(args) < 1:
-        parser.print_help()
-        sys.exit(2)
+    options_kwargs = {
+        'pre_run': options.pre_run,
+        'steal_output': options.steal_output,
+    }
 
-    mainpyfile = args[0]
+    if options.module:
+        sys.argv = args
+        from pudb import runmodule
+        runmodule(options.module, **options_kwargs)
+    else:
+        if len(args) < 1:
+            parser.print_help()
+            sys.exit(2)
 
-    from os.path import exists
-    if not exists(mainpyfile):
-        print('Error: %s does not exist' % mainpyfile)
-        sys.exit(1)
+        mainpyfile = args[0]
 
-    sys.argv = args
+        from os.path import exists
+        if not exists(mainpyfile):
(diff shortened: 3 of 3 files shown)


pudb/debugger.py
```diff
@@ -486,6 +486,25 @@ def _runscript(self, filename):
...
+            "__file__": self.mainpyfile,
+            "__package__": mod_spec.parent,
+            "__loader__": mod_spec.loader,
+            "__spec__": mod_spec,
+            "__builtins__": __builtins__,
+        })
+
+        self._wait_for_mainpyfile = True
+
+        self.run(code)
```
eliasdorneles, line 506: This bit was basically stolen from pdb's code: https://github.com/python/cpython/blob/3.8/Lib/pdb.py#L1530

pudb/run.py
```diff
@@ -5,33 +5,42 @@ def main():
...
+        if not exists(mainpyfile):
+            print('Error: %s does not exist' % mainpyfile)
+            sys.exit(1)
+
+        sys.argv = args
+
+        from pudb import runscript
+        runscript(mainpyfile,
+                pre_run=options.pre_run,
+                steal_output=options.steal_output)
```
inducer, line 43: The options processing is a bit redundant. Maybe make an `options_kwargs` above to minimize duplication? If you don't like that, I'm also OK with comments to remind folks to keep the two copies consistent.

pudb/debugger.py
inducer: Could you add the provenance as a comment?

pudb/__init__.py
```diff
@@ -132,7 +138,14 @@ def runscript(mainpyfile, args=None, pre_run="", steal_output=False):
...
         prev_sys_argv = sys.argv[:]
-        sys.argv = [mainpyfile] + args
+        if run_as_module:
+            sys.argv = args
+        else:
+            sys.argv = [mainpyfile] + args
+    else:
+        if run_as_module:
+            prev_sys_argv = sys.argv[:]
+            sys.argv = sys.argv[2:]
```
inducer, line 148: This is not necessarily correct. `OptionParser` as called below will tolerate `-m SOMETHING` in places other than the first two arguments.

pudb/run.py
eliasdorneles: Sounds good, will do!

pudb/__init__.py
eliasdorneles: Hm, I admit I've added it without fully understanding...
But when I do `pudb3 -m json.tool`, it tries to open `json.tool` if I don't do this as if it were the JSON file.
I'll do some more tests.

pudb/debugger.py
eliasdorneles: Will do!
I also just realized this breaks in PY2, will fix that as well.

pudb/__init__.py
eliasdorneles: Ahh, you're right, I did a few more testing locally and I got it now, I actually need to update `sys.argv` before calling runmodule as it's done for runscript -- will fix in a jiffy!

asmeurer:
If you try to run a module that doesn't exist, you get an error like

```pytb
  File "/Users/aaronmeurer/Documents/pudb/pudb/__init__.py", line 167, in runscript
    dbg._runmodule(mainpyfile)
  File "/Users/aaronmeurer/Documents/pudb/pudb/debugger.py", line 491, in _runmodule
    mod_name, mod_spec, code = runpy._get_module_details(module_name)
  File "/Users/aaronmeurer/anaconda3/lib/python3.7/runpy.py", line 136, in _get_module_details
    raise error("No module named %s" % mod_name)
ImportError: No module named fdjkslf

During handling of the above exception, another exception occurred:

  File "/Users/aaronmeurer/anaconda3/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/Users/aaronmeurer/anaconda3/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Users/aaronmeurer/Documents/pudb/pudb/__main__.py", line 31, in <module>
    main()
  File "/Users/aaronmeurer/Documents/pudb/pudb/run.py", line 25, in main
    steal_output=options.steal_output)
  File "/Users/aaronmeurer/Documents/pudb/pudb/__init__.py", line 125, in runmodule
    runscript(*args, **kwargs)
  File "/Users/aaronmeurer/Documents/pudb/pudb/__init__.py", line 176, in runscript
    dbg.interaction(None, sys.exc_info())
  File "/Users/aaronmeurer/Documents/pudb/pudb/debugger.py", line 389, in interaction
    self.stack, index = self.get_shortened_stack(frame, tb)
  File "/Users/aaronmeurer/Documents/pudb/pudb/debugger.py", line 350, in get_shortened_stack
    stack, index = self.get_stack(frame, tb)
  File "/Users/aaronmeurer/anaconda3/lib/python3.7/bdb.py", line 520, in get_stack
    if f is self.botframe:
AttributeError: 'Debugger' object has no attribute 'botframe'
```

I think only the top-level ImportError message should be shown. I think somewhere it needs to catch the ImportError and give up.

asmeurer:
It's a little disappointing to see that pudb is still using optparse instead of argparse, but I guess that can be fixed in a separate issue. 

I noticed that it allows more than one script argument, even though only the first one is processed. This is an issue also in master, so maybe it doesn't need to be fixed here, but when we do we should not allow any script arguments if `-m` is used (unless we instead want to support debugging one script after the other).

asmeurer:
Actually, scrap that. It is passing the "second" script as the argv, like it should.

pudb/run.py
```diff
@@ -5,33 +5,44 @@ def main():
...
 
     from optparse import OptionParser
+    # TODO: update cmd string
     parser = OptionParser(
-            usage="usage: %prog [options] SCRIPT-TO-RUN [SCRIPT-ARGUMENTS]")
+            usage="usage: %prog [options] [SCRIPT-TO-RUN|-m MODULE] [ARGUMENTS]")
 
     parser.add_option("-s", "--steal-output", action="store_true"),
+    parser.add_option("-m", "--module", metavar="MODULE",
+                      help="Debug module or package instead of script"),
```
asmeurer, line 14: Add something here about how this terminates the options list (c.f. `python -h`), since any further arguments should be passed as the argv to the script.

pudb/run.py
```diff
@@ -5,33 +5,44 @@ def main():
     import sys
 
     from optparse import OptionParser
+    # TODO: update cmd string
     parser = OptionParser(
-            usage="usage: %prog [options] SCRIPT-TO-RUN [SCRIPT-ARGUMENTS]")
+            usage="usage: %prog [options] [SCRIPT-TO-RUN|-m MODULE] [ARGUMENTS]")
```
asmeurer, line 10: This isn't related to this PR, but I noticed that if you run `python -m pudb`, the prog is set to `__main__.py`.

asmeurer:
For the ImportError, we may want to just print the error message instead showing the full traceback, like how `python -m` does.

pudb/run.py
eliasdorneles: Hm, but it doesn't really, does it?
It seems `OptionParser` will keep parsing until it doesn't recognize any other args, I can run a module with `pudb3 -m my_module --pre-run="echo oi" one two three` and OptionParser will give `--pre-run="echo oi"` to pudb and `one two three` to the module

asmeurer: I guess you're right. That's how it should behave though. I don't know if optparse or argparse actually support this.

eliasdorneles:
> For the ImportError, we may want to just print the error message instead showing the full traceback, like how `python -m` does.

Makes sense, will do!

pudb/run.py
asmeurer: Looks like with argparse you can use `REMAINDER`. I don't know if it is possible to do this with optparse. We may need to move over to argparse for this PR. If `-m` does pass all the flags to the module that will be very surprising to people. Does pdb do the right thing here?

asmeurer: You can use something like `python -m pudb -m http.server -h` as a test case. It should open the debugger in pudb, and after continuing and exiting, the http.server help should be printed to the terminal. If it shows the pudb help that behavior is wrong.

eliasdorneles: It's quite late for me, so i'll be off now.
If you leave comments, I can look tomorrow.

asmeurer: Looks like pdb parses the arguments manually with getopt https://github.com/python/cpython/blob/e822e37946f27c09953bb5733acf3b07c2db690f/Lib/pdb.py#L1661. I don't see why argparse.REMAINDER wouldn't work, but if there is some issue with it, I guess we can use that. That will make it hard if we want to add any other command line arguments in the future, so we should try to use argparse if we can.

asmeurer:
> Makes sense, will do!

Be sure to test both cases: a module that doesn't exist, and a package that does exist but doesn't have `__main__.py`.

pudb/run.py
eliasdorneles: It doesn't seem to be possible to get the behavior you want for `python -m pudb -m http.server -h` using argparse.
Here is a test script that demonstrates the behavior, which you can try yourself: https://gist.github.com/eliasdorneles/525991dbc996b59ef862a1650fce00eb

If you run it like: `python script.py -m module -h` you will always get the script's help.
In order to get the `-h` passed on to the `script_args` variable, you have to either add another argument before the `-h` or add `--` (which is pretty reasonable to me, that's how many Unix utilities handle these cases of partial parsing).

If this sounds good enough, I can go ahead and change the code to use argparse.

eliasdorneles:
> Be sure to test both cases: a module that doesn't exist, and a package that does exist but doesn't have `__main__.py`.

Just did, in both cases it works the same as `python -m`.

pudb/run.py
eliasdorneles: Personally, I believe this PR is already useful and making things better, despite of the argument handling bikeshedding. :angel:

asmeurer: Presumably that's only because -h is also a flag to pudb. So any other flag would still be passed through. If that's the case, that's better than the current behavior, which steals all `--flags`. 

Requiring for `--` for the corner cases of flags used by pudb is probably fine, though it should be documented in the help. It does mean that we should minimize the single letter flags to pudb to avoid conflicts, but I don't think that's a huge deal. Making it not required would be better but that may be too complicated to implement. 

> Personally, I believe this PR is already useful and making things better, despite of the argument handling bikeshedding. 👼

I don't see it as bikeshedding. -m modules are very often run with command line flags, so users would run into the bug of the current behavior quite quickly I think.

eliasdorneles: Fair enough.

Well, I thought since pudb doesn't have many arguments, another option is to write our own parsing code that does exactly what we want and doesn't introduce any of the limitations of `getopt` and no ambiguities of using `optparse`/`argparse`.

Here is what it would like: https://gist.github.com/eliasdorneles/51fc37dd4910fe07694a30ed6a70aa57

Would it be acceptable to do that here?

asmeurer: How is that different from using getopt? I haven't really used getopt before so I don't know much about it.

asmeurer: Maybe an easier way is to manually extract everything before `-m` in the sys.argv and pass only that to `parse_args()`. We could even keep using `optparse` with this method (if you don't want to translate to argparse here).

eliasdorneles: Ah, my bad, I had assumed getopt was limited to single-character flags (I had it mixed up with the `getopts` bash builtin).
I will look into the getopt way, as it seems simple enough, less workaround-ish (I'm afraid splitting on `-m` would introduce annoying edge cases, e.g, if it was meant to be passed to a script) and it seems a good idea to mirror the `pdb` way.

eliasdorneles: So I looked into it more closely, and the thing with `getopt` is that it consumes all arguments until it finds a non-argument (much like `parse_known_args` from argparse), and the reason why `python -m pdb -m http.server -h` works with it is that the `-m` is implemented as a boolean flag -- meaning that getopt stops parsing arguments when it see `http.server` because it's the first non-argument.

You can see this by running `python -m pdb -m -c continue http.server -h` <- notice how the -m is given (to activate mode "run as module") and then the module name is given only after given another option recognized by pdb (`-c continue`).

So, if we're okay using this same behavior of `-m` as a boolean flag on pudb, we can then easily implement it with `argparse` and `argparse.REMAINDER`.

I'm going to push a commit implementing that, let me know what do you think.

asmeurer: > `python -m pdb -m -c continue http.server -h`

That seems like a corner case that isn't that important. If that's the only thing that is off by that implementation, I am favor of it.

eliasdorneles: Great! I agree!
I've pushed the implementation with argparse, it seems to be working well.
Does it look good now?
Thanks!

asmeurer:
I've tested the new code. It seems to work. I noticed that if you pass it -m but give it a file instead of a module, it just executes the file without running the debugger. I'm not sure if this is intentional, although it does seem to match `python -m` and `pdb -m` behavior.

eliasdorneles:
> I've tested the new code. It seems to work. I noticed that if you pass it -m but give it a file instead of a module, it just executes the file without running the debugger. I'm not sure if this is intentional, although it does seem to match `python -m` and `pdb -m` behavior.

Oh, I hadn't tested that case. Well, it seems okay to leave it like that?

pudb/run.py
```diff
@@ -4,34 +4,45 @@
...
-    runscript(mainpyfile,
-            pre_run=options.pre_run,
-            steal_output=options.steal_output)
+    if options.module:
+        from pudb import runmodule
+        runmodule(mainpyfile, **options_kwargs)
+    else:
+        from os.path import exists
+        if not exists(mainpyfile):
+            print('Error: %s does not exist' % mainpyfile)
```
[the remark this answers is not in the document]
eliasdorneles: Makes sense, I'll change that -- it changes current behavior though. :upside_down_face:

pudb/run.py
```diff
@@ -4,34 +4,45 @@
...
-            usage="usage: %prog [options] SCRIPT-TO-RUN [SCRIPT-ARGUMENTS]")
-
-    parser.add_option("-s", "--steal-output", action="store_true"),
-    parser.add_option("--pre-run", metavar="COMMAND",
+    import argparse
+    parser = argparse.ArgumentParser(
+        usage="%(prog)s [options] [-m] SCRIPT-OR-MODULE-TO-RUN [SCRIPT_ARGS]"
+    )
+    parser.add_argument("-s", "--steal-output", action="store_true"),
+    parser.add_argument("-m", "--module", action='store_true',
```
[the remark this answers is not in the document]
eliasdorneles: Sure, I will add a comment!

However, I think it's best not to mention it in the help, I believe we should consider it as an implementation detail (i.e. people shouldn't rely that it should stay like this, and the fact that `-m -s --pre-run STUFF mymodule` works is just because "Simple is better than complex" and "practicality beats purity"). =)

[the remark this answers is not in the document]
eliasdorneles: Done! Back to you :)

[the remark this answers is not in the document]
asmeurer: Yeah this -m stuff should definitely be mentioned in a code comment. I don't know if the implementation details need to be outlined in the help. Stuff like `-m -s --pre-run STUFF mymodule` isn't actually supported. It just works because the only reasonable way to implement it with argparse requires it to be implemented that way.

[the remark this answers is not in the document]
asmeurer: And I guess the comment in question is already there. So this looks fine to me (just from looking at it. I didn't test the latest code).

inducer:
Cool. Thanks @asmeurer for taking another look. In it goes.

inducer:
And thanks @eliasdorneles for your contribution!

eliasdorneles:
Awesome, thank you both!

asmeurer:
Thanks for taking the time to work through the issues here. And I'm also happy to see the command line code now use argparse instead of optparse.

delbao11:
do we have the latest release version with this PR merged?

inducer:
Thanks for the reminder. 2020.1, out now.

PR #391, opened 2020-06-23
Drop py2

inducer:


setup.py
```diff
@@ -22,6 +22,7 @@
       long_description=long_description,
       author='Andreas Kloeckner',
       author_email='inform@tiker.net',
+      python_requires="~=3.6",
```
asmeurer, line 25: What is `~=`?

inducer: https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires

asmeurer: OK. Seems like an odd thing considering there will never be a Python 4 ;)

asmeurer:
I don't know if you want to remove any compatibility code. 

Otherwise this looks fine. I think the main thing is the python_requires in the setup.py so that Python 2 pip will not try to install this version.

inducer:
> I don't know if you want to remove any compatibility code.

I thought about that, but I decided against the big crusade for now. That can either happen along the way or in one big batch later.

> Otherwise this looks fine.

Thanks for taking a look! I'll merge it then.

PR #392, opened 2020-06-28
Map up/down keys to command history prev/next

### LINKED ISSUE #348: Command history with cursor up in internal shell
When inside the internal shell the cursor-up key just selects the lines in history with no (apparent) useful function. 

It would be nice if it could instead access the history (as does ctrl-p/n).

ned2:
Map up/down keys to command history prev/next when command line is focused

Fixes #348

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -96,7 +96,7 @@ Shell-related:
     _/= - minimize/maximize inline shell (active in command line history)
 
     Ctrl-v - insert newline
-    Ctrl-n/p - browse command line history
+    Arrow Down/Up - browse command line history
     Tab - yes, there is (simple) tab completion
 """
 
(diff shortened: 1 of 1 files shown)


asmeurer:
It's not obvious, but you can still arrow up through the history items with this by pressing "right" to select the "<clear>" button, then arrowing up.

inducer:
- Fix the help
- Could you see if we can bind shift+up (or some such) to paging through the console scrollback?

ned2:
Oops yes the help, I forgot about that. done.

With the console scrollback paging, @inducer, do you mean navigating forward/back through blocks of history commands of size greater than 1, or something else?

inducer:
> With the console scrollback paging, @inducer, do you mean navigating forward/back through blocks of history commands of size greater than 1, or something else?

Just navigating *somehow*. Since this takes away cursor up/down as a canonical way to do scrollback navigation, I'd like for something somewhat easy to remember to be put in place. And @asmeurer's go-right-then-up thing is cute, but doesn't count: it's not convenient, obvious, or easy-to-remember.

sisrfeng:
any update?

> Agree, this is super annoying. It's really messing up my muscle memory.
I Agree!

asmeurer:
I think if these two things are addressed this is fine https://github.com/inducer/pudb/pull/392#issuecomment-650694840. I would suggest making a new PR for this, since no work has happened on this one in a while.

inducer:
This was done in #600.

PR #396, opened 2020-08-15
Use h/l in the variable to view to collapse/open the current container

mvanderkamp:
This behaves similarly to, for example, tmux's interactive session/window/pane selection.
I did see that h/l were supposed to be mapped to movement in the RHS column already, but this doesn't seem to work for me, and I believe this is a more useful interaction anyway. That said, I'm happy to move it this functionality to a different key.

The benefit is that you can open a huge container, navigate down through it, then when you're done you don't have to scroll all the way back up; just slap 'h'.

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -527,7 +527,7 @@ class Debugger(bdb.Bdb):
 from pudb.ui_tools import make_hotkey_markup, labelled_value, \
         SelectableText, SignalWrap, StackFrame, BreakpointFrame
 
-from pudb.var_view import FrameVarInfoKeeper, ValueWalker
+from pudb.var_view import FrameVarInfoKeeper
 
 
 # {{{ display setup
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -29,6 +29,7 @@ THE SOFTWARE.
 """
 
 
+import __builtin__
 import urwid
 import bdb
 import gc
@@ -525,7 +526,7 @@ class Debugger(bdb.Bdb):
 from pudb.ui_tools import make_hotkey_markup, labelled_value, \
         SelectableText, SignalWrap, StackFrame, BreakpointFrame
 
-from pudb.var_view import FrameVarInfoKeeper
+from pudb.var_view import FrameVarInfoKeeper, ValueWalker
 
 
 # {{{ display setup
@@ -710,6 +711,7 @@ class DebuggerUI(FrameVarInfoKeeper):
         self.search_controller = SearchController(self)
 
         self.last_module_filter = ""
+        self.var_focus_index = None
 
         # {{{ build ui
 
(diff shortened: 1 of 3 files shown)


asmeurer:
It seems reasonable to me. Actually h/l scroll in the code view, so I don't think they are supposed to move between panes.

mvanderkamp:
Ah okay I wasn't sure. I just saw the potential conflict and wanted to bring it up.

pudb/debugger.py
```diff
@@ -878,17 +881,46 @@ def change_rhs_box(name, index, direction, w, size, key):
...
+                        for variable_widget in reversed(self.locals[:pos+1])
+                        if variable_widget.prefix == desired_prefix
+                    )
+                except StopIteration:
+                    # No parent found, so don't do anything
+                    pass
+                else:
+                    p_iinfo = get_inspect_info(parent.id_path)
+                    p_iinfo.show_detail = False
+                    self.var_focus_index = self.locals.index(parent)
```
inducer, line 908: I agree that this works currently, but I don't like that this uses a presentation-related property (`prefix`) for logic purposes (as there's no guarantee that this string will always look the way it does now). Could you refactor this/the `VariableWidget` so that the `VariableWidget` carries a (structural) `nesting_level` attribute instead (and, ideally, compute the prefix from that)?

pudb/debugger.py
```diff
@@ -878,17 +881,46 @@ def change_rhs_box(name, index, direction, w, size, key):
...
+                try:
+                    # Walk backwards to find first shorter prefix
+                    parent = next(
+                        variable_widget
+                        for variable_widget in reversed(self.locals[:pos+1])
+                        if variable_widget.prefix == desired_prefix
+                    )
+                except StopIteration:
+                    # No parent found, so don't do anything
+                    pass
```
inducer, line 904: Oh, one more thing: What do you think of popping up an error message here instead of silently swallowing the key? I'm not sure it's needed, but I figured I'd raise the question.

pudb/debugger.py
mvanderkamp: Yup I'm happy to do so. I initially avoided this approach to try to minimize the impact of my patch but I agree this would be better!

pudb/debugger.py
mvanderkamp: Hmm... This case will be hit at the top level only. We don't give the user an error message when they do other similar ineffective operations, (e.g. pressing 'up' when they're already at the top of the variable list, or 'down' from the bottom of the breakpoint list), so I think I'd prefer to not popup an error.

That said, if 'h' is supposed to have some default behaviour across the right hand column, would returning the "key" variable at that point allow the default behaviour to trigger?

inducer: That's fair. I'm OK with swallowing the key. Having it do something different (not that the key is currently mapped) depending on context would be confusing IMO. Thanks for considering it.

mvanderkamp:
I decided to take it one step further and have each VariableWidget store a reference to its parent. That way we don't have to search through the widget list at all, but can just grab the reference directly. I did go with storing the nesting level as well though, as it does make generating the prefix easier.



Given the changes and fixups, I'm happy to do a rebase if you want.

pudb/debugger.py
```diff
@@ -710,6 +712,7 @@ def __init__(self, dbg, stdin, stdout, term_size):
         self.search_controller = SearchController(self)
 
         self.last_module_filter = ""
+        self.var_focus_index = None
```
inducer, line 715: Not a fan of this being global state. As far as I can tell, it's only being used to get the index from `collapse_current` into `update_var_view`. I would much prefer if the lifetime of this state were limited to the subroutine calling both `change_var_state`. If it hangs around longer, it just produces questions regarding its validity.

inducer:
> Given the changes and fixups, I'm happy to do a rebase if you want.

Thanks for offering to do that. I'm not much of a history gardener, so it's not needed from my point of view.

pudb/debugger.py
mvanderkamp: Done! Thanks for pointing that out, I hadn't spotted that there was an easy way to connect these two.

inducer:
Just played with this a bit, and I like it a lot. I also like how the last iteration of the code came out. Thanks for your contribution!

Issue #397, opened 2020-08-15
Help text always shows the "side" help


mvanderkamp:
The "main" help text with the source view help first is no longer shown when pressing '?' from the source view.

PR #398, opened 2020-08-15
Fix issue #397: show "main" help if in source view

mvanderkamp:
Looks like this was missed in the refactor in commit 86443c7818c0f67dfc70b4aa45eb609f485e256f


inducer:
Thanks for noticing that and submitting a fix!

PR #400, opened 2020-08-18
Use isroutine instead of ismethod to detect functions

mvanderkamp:
ismethod misses functions whose implementation is not in python. We could split hairs and provide another action for toggling whether non-python/builtin functions on a given variable are visible, but I propose that this is easier, and probably what the user usually wants to do anyway.

An example where this would be useful is working with PySide (Qt for python).


inducer:
LGTM. Thanks! (Also, TIL about `isroutine`.)

Issue #28, opened 2011-12-28
Enable Remote Debugging


aventurella:
Per Issue:
http://lists.tiker.net/pipermail/pudb/2010-December/000000.html

I would be great to have remote debugging enabled to facilitate web application debugging. Where the debugger is initialized in the application and then pudb is able to attach to the process.

gasull:
Why is this still open if rudimentary [remote debugging is already supported](https://documen.tician.de/pudb/starting.html#remote-debugging)?

inducer:
Thanks for noticing this! I think a fair answer is that this should be closed, and new, more focused issues be opened with ideas for improvement of the current remote debugging code.

Issue #286, opened 2017-12-25
Cannot use pudb for multiprocessing


xuananh-viosoft:
I am trying to use pudb to debug with multiprocessig, but I encounter error as below:
Code:

```
import pudb
import multiprocessing
import time

def worker():
    i = 0
    while i < 10:
        pudb.set_trace()
        i = i + 1
        time.sleep(1)
    
if __name__ == '__main__':
    p1 = multiprocessing.Process(target=worker)
    p1.start()
```

Error:

> Process Process-1:
> Traceback (most recent call last):
>   File "/usr/lib/python2.7/multiprocessing/process.py", line 258, in _bootstrap
>     self.run()
>   File "/usr/lib/python2.7/multiprocessing/process.py", line 114, in run
>     self._target(*self._args, **self._kwargs)
>   File "pudb_sample.py", line 9, in worker
>     i = i + 1
>   File "pudb_sample.py", line 9, in worker
>     i = i + 1
>   File "/usr/lib/python2.7/bdb.py", line 49, in trace_dispatch
>     return self.dispatch_line(frame)
>   File "/usr/local/lib/python2.7/dist-packages/pudb/debugger.py", line 160, in dispatch_line
>     self.user_line(frame)
>   File "/usr/local/lib/python2.7/dist-packages/pudb/debugger.py", line 381, in user_line
>     self.interaction(frame)
>   File "/usr/local/lib/python2.7/dist-packages/pudb/debugger.py", line 349, in interaction
>     show_exc_dialog=show_exc_dialog)
>   File "/usr/local/lib/python2.7/dist-packages/pudb/debugger.py", line 2089, in call_with_ui
>     return f(*args, **kwargs)
>   File "/usr/local/lib/python2.7/dist-packages/pudb/debugger.py", line 2330, in interaction
>     self.event_loop()
>   File "/usr/local/lib/python2.7/dist-packages/pudb/debugger.py", line 2274, in event_loop
>     % VERSION)
>   File "/usr/local/lib/python2.7/dist-packages/pudb/debugger.py", line 1927, in message
>     [("OK", True)], title=title, **kwargs)
>   File "/usr/local/lib/python2.7/dist-packages/pudb/debugger.py", line 2089, in call_with_ui
>     return f(*args, **kwargs)
>   File "/usr/local/lib/python2.7/dist-packages/pudb/debugger.py", line 2008, in dialog
>     return self.event_loop(w)[0]
[...]

PR #402, opened 2020-08-28
add debugger for forked processes

jstriebel:
This PR adds a version of pudb that can be used from forked processes without using telnet. 

related to #28

Thanks for considering this PR! I'm happy to adapt anything if needed.

### DIFF
--- a/doc/starting.rst
+++ b/doc/starting.rst
@@ -81,7 +81,7 @@ connection::
     pudb:6899: Waiting for client...
 
 Using the debugger after forking
-^^^^^^^^^^^^^^^^^^^^^^^^^^
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 In a forked process no tty is attached usually, which leads to
 ``TypeError: ord() expected a character, but string of length 0 found``
--- a/doc/starting.rst
+++ b/doc/starting.rst
@@ -101,8 +101,7 @@ Running it with standard pudb breaks::
     PYTHONBREAKPOINT=pudb.set_trace python script.py
 
 However, on Unix system, e.g. Linux & MacOS, debugging a forked
-process is supported using ``pudb.forked.set_trace``, which redirects
-in- and output to ``/dev/stdin`` & ``/dev/stdout``::
+process is supported using ``pudb.forked.set_trace``::
 
     PYTHONBREAKPOINT=pudb.forked.set_trace python script.py
 
--- a/doc/starting.rst
+++ b/doc/starting.rst
@@ -80,21 +80,33 @@ connection::
     pudb:6899: Please telnet into 127.0.0.1 6899.
     pudb:6899: Waiting for client...
 
-Debugging forked processes
+Using the debugger after forking
 ^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-On Unix system, e.g. Linux & MacOS e.g. Linux & MacOS, debugging a forked
-process is supported::
+In a forked process no tty is attached usually, which leads to
+``TypeError: ord() expected a character, but string of length 0 found``
+when debugging with standard pudb. E.g. consider this ``script.py``::
 
-    from pudb.forked import set_trace
-    set_trace()
+    from multiprocessing import Process
+    def f(name):
+        # breakpoint was introduced in Python 3.7
+        breakpoint()
+        print('hello', name)
     
-Or use it with Python's built-in ``breakpoint()`` (from Python 3.7)::
+    p = Process(target=f, args=('bob',))
+    p.start()
+    p.join()
 
-    export PYTHONBREAKPOINT=pudb.forked.set_trace
-    your-command  # using breakpoint() in the code
+Running it with standard pudb breaks::
 
-In- and output will be redirected to ``/dev/stdin`` & ``/dev/stdout``.
+    PYTHONBREAKPOINT=pudb.set_trace python script.py
+
+However, on Unix system, e.g. Linux & MacOS, debugging a forked
+process is supported using ``pudb.forked.set_trace``, which redirects
+in- and output to ``/dev/stdin`` & ``/dev/stdout``::
+
+    PYTHONBREAKPOINT=pudb.forked.set_trace python script.py
+
 
 Usage with pytest
 ^^^^^^^^^^^^^^^^^
--- a/doc/starting.rst
+++ b/doc/starting.rst
@@ -80,6 +80,22 @@ connection::
     pudb:6899: Please telnet into 127.0.0.1 6899.
     pudb:6899: Waiting for client...
 
+Debugging forked processes
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+On Unix system, e.g. Linux & MacOS e.g. Linux & MacOS, debugging a forked
+process is supported::
+
+    from pudb.forked import set_trace
+    set_trace()
+    
+Or use it with Python's built-in ``breakpoint()`` (from Python 3.7)::
+
+    export PYTHONBREAKPOINT=pudb.forked.set_trace
+    your-command  # using breakpoint() in the code
+
+In- and output will be redirected to ``/dev/stdin`` & ``/dev/stdout``.
+
 Usage with pytest
 ^^^^^^^^^^^^^^^^^
 
--- a/doc/starting.rst
+++ b/doc/starting.rst
@@ -83,7 +83,7 @@ connection::
 Using the debugger after forking
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-In a forked process no TTY is usually attached to stdin/stdout, which leads to errors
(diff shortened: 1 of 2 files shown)


pudb/forked.py
```diff
@@ -0,0 +1,30 @@
...
+            "hh",
+            fcntl.ioctl(1, termios.TIOCGWINSZ, "1234"),
+        )
+        term_size = (s[1], s[0])
+    except Exception:
+        term_size = (80, 24)
+
+    Debugger(
+        stdin=open("/dev/stdin"),
+        stdout=open("/dev/stdout", "w"),
```
inducer, line 28: Linux-specific?

pudb/forked.py
```diff
@@ -0,0 +1,30 @@
...
+
+import sys
+import fcntl
+import termios
+import struct
+
+from pudb.debugger import Debugger
+
+
+def forked_pudb():
```
inducer, line 14: Add some documentation? Maybe as a docstring?

pudb/forked.py
jstriebel: Yes, absolutely. Unfortunately I don't have much experience on Windows. I guess it should work on MacOS, too, but can't test it myself (same for Windows). I'll add it to the documention.

asmeurer: I'm not sure if PuDB works at all on Windows. We should make it work on Mac, though.

asmeurer:
I'm also confused when would you call `forked_pudb`? Should it accept the same arguments as `set_trace`? Should there be an `if __name__ == '__main__'` in the file?

jstriebel:
Sorry for the late reply. I already renamed the function to `set_trace`, since this seems to be a convention. It could accept arguments for the terminal-size and frame, similar to the remote-debugger, and also for `paused`. Should I add those?

Do you think the documentation is useful? Anything missing?

jstriebel:
> Should there be an if __name__ == '__main__' in the file?

I only started the forked debugger from the code so far, but if it makes sense as a script, sure. For now I think it already is useful to be used via a call or `breakpoint`, that's why I'd leave it as-is, similar to `remote.py`.

asmeurer:
Yes, it should accept the same arguments as set_trace, assuming they work.

jstriebel:
I added all the arguments as proposed. Anything else missing?

doc/starting.rst
```diff
@@ -80,6 +80,22 @@ connection::
     pudb:6899: Please telnet into 127.0.0.1 6899.
     pudb:6899: Waiting for client...
 
+Debugging forked processes
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+On Unix system, e.g. Linux & MacOS e.g. Linux & MacOS, debugging a forked
+process is supported::
```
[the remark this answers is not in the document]
jstriebel: When running some parts of the code in a different process, e.g. via python's multiprocessing, the tty is not attached anymore. This leads to errors for pdb and standard pudb. E.g. consider this `script.py`:
```
from multiprocessing import Process
def f(name):
    breakpoint()
    print('hello', name)

p = Process(target=f, args=('bob',))
p.start()
p.join()
```

Running `PYTHONBREAKPOINT=pudb.set_trace python script.py` results in `TypeError: ord() expected a character, but string of length 0 found` due to the missing tty. However, using `PYTHONBREAKPOINT=pudb.forked.set_trace python script.py` works in this case, since the input and output are attached explicitly. Using the `remote` pudb version would also work, but requires you to telnet into the debugger, whereas this solution directly opens the debugger on the commandline used to start python.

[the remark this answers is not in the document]
jstriebel: Just to highlight that this seems to be a use-case for more people:
There's a [medium article](https://medium.com/@auro_227/use-pudb-to-debug-python-multiprocessing-code-c0c5551d010c) about exactly this use-case, proposing `pudb.remote`.
Also a [stackoverflow question](https://stackoverflow.com/questions/47967490/cannot-use-pudb-for-multiprocessing), and an old issue (https://github.com/inducer/pudb/issues/286).

And even more discussion when looking for any python debugger that is able to debug forked processes, e.g. [here](https://stackoverflow.com/questions/4716533/how-to-attach-debugger-to-a-python-subproccess/23654936#23654936)

pudb/forked.py
inducer: @asmeurer Do you have a though on how we could achieve the same thing on a Mac?

doc/starting.rst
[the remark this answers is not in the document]
inducer: TIL that this was a problem. That (and your general approach) make sense to me now. It would be fantastic if you could improve the docs a bit to make it a bit more obvious what problem is being solved, for example by incorporating the example you supplied. Also, I'd suggest using the subsection title "Using the debugger after forking" and describing that stdin/stdout are not usually available then.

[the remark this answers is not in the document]
inducer: (And thanks for taking the time to explain!)

pudb/forked.py
jstriebel: I'd assume this works on MacOS as-is, `/dev/stdin` and `/dev/stdout` are valid there, too.

doc/starting.rst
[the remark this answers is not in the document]
jstriebel: Sure thing :-) I tried to incorporate the example. Feel free to directly add suggestions if you see more room for improvement :+1:

pudb/forked.py
inducer: OIC. LGTM then.

pudb/forked.py
```diff
@@ -0,0 +1,38 @@
...
+                "hh",
+                fcntl.ioctl(1, termios.TIOCGWINSZ, "1234"),
+            )
+            term_size = (s[1], s[0])
+        except Exception:
+            term_size = (80, 24)
+
+    Debugger(
+        stdin=open("/dev/stdin"),
+        stdout=open("/dev/stdout", "w"),
```
inducer, line 36: Should these be opened in binary mode?

doc/starting.rst
```diff
@@ -80,6 +80,34 @@ connection::
     pudb:6899: Please telnet into 127.0.0.1 6899.
     pudb:6899: Waiting for client...
 
+Using the debugger after forking
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+In a forked process no tty is attached usually, which leads to
+``TypeError: ord() expected a character, but string of length 0 found``
```
inducer, line 87:
```suggestion
In a forked process no TTY is usually attached to stdin/stdout, which leads to errors
```

doc/starting.rst
```diff
@@ -80,6 +80,34 @@ connection::
...
+    p.start()
+    p.join()
+
+Running it with standard pudb breaks::
+
+    PYTHONBREAKPOINT=pudb.set_trace python script.py
+
+However, on Unix system, e.g. Linux & MacOS, debugging a forked
+process is supported using ``pudb.forked.set_trace``, which redirects
+in- and output to ``/dev/stdin`` & ``/dev/stdout``::
```
inducer, line 106:
```suggestion
process is supported using ``pudb.forked.set_trace``::
```

inducer (commented):
I'm just about happy with this. Just a few minor wrinkles left. And thanks for your patience in dealing with my comments!

pudb/forked.py
jstriebel: I wasn't sure and tested it, but using binary mode results in an `io.UnsupportedOperation: File or stream is not seekable.` error.

doc/starting.rst
```diff
@@ -80,6 +80,32 @@ connection::
...
+    
+    p = Process(target=f, args=('bob',))
+    p.start()
+    p.join()
+
+Running it with standard pudb breaks::
+
+    PYTHONBREAKPOINT=pudb.set_trace python script.py
+
+However, on Unix system, e.g. Linux & MacOS, debugging a forked
```
inducer, line 103:
```suggestion
However, on Unix systems, e.g. Linux & MacOS, debugging a forked
```

doc/starting.rst
```diff
@@ -80,6 +80,32 @@ connection::
     pudb:6899: Please telnet into 127.0.0.1 6899.
     pudb:6899: Waiting for client...
 
+Using the debugger after forking
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+In a forked process no TTY is usually attached to stdin/stdout, which leads to errors
```
inducer, line 86:
```suggestion
In a forked process, no TTY is usually attached to stdin/stdout, which leads to errors
```

pudb/forked.py
inducer: Hmm, thanks for checking. I'm not sure I understand the failure mode, but I'm OK with leaving this be for now.

inducer:
Thanks for your contribution! This is ready to go in once CI passes.

Issue #355, opened 2019-07-24
config reset


hovnatan:
Sometimes the config file at `~/.config/pudb/pudb.cfg` gets reset to the default value. Using `pudb 2019.1`

inducer:
I haven't had that happen. Do you know of a way to reproduce this behavior?

hovnatan:
Please run the attached code with `export PYTHONBREAKPOINT=pudb.remote.set_trace`. Then `telnet` into one of the connections. This resets the `pudb.cfg`
```
from torch.utils.data import DataLoader, Dataset

class MyDataset(Dataset):
    def __getitem__(self, i):
        breakpoint()
        return i

    def __len__(self):
        return 1000000


mydata = MyDataset()
myloader = DataLoader(mydata, batch_size=64, num_workers=16)

it = iter(myloader)
item = next(it)
print(item)
```

inducer:
I tried with a different file (I don't have PyTorch installed), and my configuration survived without a problem.

inducer:
Maybe PyTorch is doing something that it shouldn't?

asmeurer:
Right now save_config swallows all exceptions https://github.com/inducer/pudb/blob/3f041b88b7cf623a22d50e702663b1e3e8076f3c/pudb/settings.py#L150. Can you try removing the `except Exception` and seeing if an exception occurs. Ditto with load_config.

asmeurer:
My guess is that load_config fails on loading the file, but doesn't do anything because of an `except Exception: pass`.  So it sets everything as the default, then immediately saves it: https://github.com/inducer/pudb/blob/3f041b88b7cf623a22d50e702663b1e3e8076f3c/pudb/debugger.py#L44-L45

hovnatan:
This is what I get if I comment out `except Exception: ... `
```
> /home/hovnatan/bug/temp.py(7)__getitem__()
-> return i
(Pdb) > /home/hovnatan/bug/temp.py(7)__getitem__()
-> return i
> /home/hovnatan/bug/temp.py(7)__getitem__()
-> return i

(Pdb) (Pdb)

> /home/hovnatan/bug/temp.py(7)__getitem__()
-> return i
(Pdb)
Traceback (most recent call last):
  File "temp.py", line 17, in <module>
    item = next(it)
  File "/home/hovnatan/miniconda3/lib/python3.7/site-packages/torch/utils/data/dataloader.py", line 582, in __next__
[...]

PR #404, opened 2020-09-06
Provide an option to log unexpected exceptions from pudb to stderr

mvanderkamp:
It's not much, but it's a start towards making weird issues that are currently swallowed into the abyss a bit easier to debug. The loggers are set up to not write anything out by default, they're just there as something that can be turned on when we want to figure out why something strange is happening.

The logging calls themselves are a bit arbitrary and preliminary in this PR, the main thing I think is having a logging structure in place to begin with for these sorts of situations. Apologies if one already exists that I managed to overlook!

Very much open to feedback on this though, I wasn't sure what direction to go in so this is just a stab in the dark at something that could be useful.

On a related note, I think I'd like to take a pass at cleaning up settings.py, I think there's some odd behaviour in terms of where configs get loaded/saved to, and I think it would be nice if someone using this code as an API could specify a filename for saving/loading configs and/or breakpoint files. But that's for another day...

This is particularly in response to issue #355

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -2613,7 +2613,9 @@ class DebuggerUI(FrameVarInfoKeeper):
                     class_name = frame.f_locals["self"].__class__.__name__
                 except Exception:
                     from pudb.lowlevel import ui_log
-                    ui_log.exception('Failed to determine class name')
+                    message = 'Failed to determine class name'
+                    ui_log.exception(message)
+                    class_name = '!! %s !!' % message
 
             return StackFrame(frame is self.debugger.curframe,
                     code.co_name, class_name,
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -1595,6 +1595,7 @@ class DebuggerUI(FrameVarInfoKeeper):
             self.cmdline_list.set_focus_valign("bottom")
             self.cmdline_list.set_focus(len(self.cmdline_contents) - 1,
                     coming_from="above")
+        self.add_cmdline_content = add_cmdline_content
 
         def cmdline_tab_complete(w, size, key):
             try:
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -1583,25 +1583,11 @@ class DebuggerUI(FrameVarInfoKeeper):
                     [curframe.f_locals, curframe.f_globals],
                     curframe.f_locals)
 
-        def add_cmdline_content(s, attr):
-            s = s.rstrip("\n")
-
-            from pudb.ui_tools import SelectableText
-            self.cmdline_contents.append(
-                    urwid.AttrMap(SelectableText(s),
-                        attr, "focused "+attr))
-
-            # scroll to end of last entry
-            self.cmdline_list.set_focus_valign("bottom")
-            self.cmdline_list.set_focus(len(self.cmdline_contents) - 1,
-                    coming_from="above")
-        self.add_cmdline_content = add_cmdline_content
-
         def cmdline_tab_complete(w, size, key):
             try:
                 from jedi import Interpreter
             except ImportError:
-                add_cmdline_content(
+                self.add_cmdline_content(
                         "Tab completion requires jedi to be installed. ",
                         "command line error")
                 return
@@ -1609,7 +1595,7 @@ class DebuggerUI(FrameVarInfoKeeper):
             import jedi
             from distutils.version import LooseVersion
             if LooseVersion(jedi.__version__) < LooseVersion("0.16.0"):
-                add_cmdline_content(
+                self.add_cmdline_content(
                         "jedi 0.16.0 is required for Tab completion",
                         "command line error")
 
@@ -1625,7 +1611,7 @@ class DebuggerUI(FrameVarInfoKeeper):
                         [cmdline_get_namespace()]).complete()
             except Exception as e:
                 # Jedi sometimes produces errors. Ignore them.
-                add_cmdline_content(
+                self.add_cmdline_content(
                         "Could not tab complete (Jedi error: '%s')" % e,
                         "command line error")
                 return
@@ -1656,7 +1642,7 @@ class DebuggerUI(FrameVarInfoKeeper):
             if (
                     len(completed_chopped_text) == 0
                     and len(completions) > 1):
-                add_cmdline_content(
+                self.add_cmdline_content(
                         "   ".join(full_completions),
                         "command line output")
                 return
@@ -1676,7 +1662,7 @@ class DebuggerUI(FrameVarInfoKeeper):
                 # blank command -> refuse service
                 return
 
-            add_cmdline_content(">>> " + cmd, "command line input")
+            self.add_cmdline_content(">>> " + cmd, "command line input")
 
             if not self.cmdline_history or cmd != self.cmdline_history[-1]:
                 self.cmdline_history.append(cmd)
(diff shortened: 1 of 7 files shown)


mvanderkamp:
I should mention- one nice thing about the approach of just logging to stderr is it makes it easy for users to do things like:
```
python -m pudb --log-errors debug_me.py 2> errors.log
```
And then in another shell:
```
tail -f errors.log
```

pudb/var_view.py
```diff
@@ -456,7 +465,7 @@ def walk_value(self, parent, label, value, id_path=None, attr_prefix=None):
             try:
                 key_its.append(dir(value))
             except Exception:
-                pass
+                ui_log.exception('Failed to lookup attributes')
```
inducer, line 468:
```suggestion
                ui_log.exception('Failed to look up attributes')
```

pudb/lowlevel.py
```diff
@@ -28,6 +28,26 @@
...
+ui_log.addHandler(ui_handler)
+
+settings_formatter = logging.Formatter(
+    fmt='\n*** Pudb Settings Exception Encountered: %(message)s ***\n'
+)
+settings_handler = logging.StreamHandler()
+settings_handler.setFormatter(settings_formatter)
+settings_log = logging.getLogger('settings')
+settings_log.setLevel(logging.CRITICAL)
+settings_log.addHandler(settings_handler)
```
inducer, line 49: > I should mention- one nice thing about the approach of just logging to stderr is it makes it easy for users to do things like:

That's neat. Could you add a snippet to the docs about that to make that discoverable?

inducer (commented):
Thanks! This is nice. A few comments below.

asmeurer:
Does logging to stderr break the UI? What's the simplest way to test this?

pudb/lowlevel.py
mvanderkamp: Done!

inducer:
This LGTM. Thanks for your work on this @mvanderkamp! @asmeurer, any objections?

asmeurer:
I didn't get a chance to test this, but as long as it doesn't break the UI it looks fine to me.

mvanderkamp:
> Does logging to stderr break the UI? What's the simplest way to test this?

The log does interfere with the UI, which is why I disabled it by default and recommended piping to a file. If you'd prefer, I could require a path argument to the log option and write the logs directly to that file instead. I'm not so familiar with how the various file descriptors are handled here yet, so I wasn't sure of a better approach.

A simple way to try it out is to put something that will raise an exception into a `safely_stringify_for_pudb` method of whatever you're debugging.

asmeurer:
What about disabling the UI like it does when stepping through code?

mvanderkamp:
Do you mean show it as a popup? Or explicitly show the output stream, require confirmation, then continue?

asmeurer:
Right now when you press `n` or `c`, it disables the UI so that any output will just go to the normal terminal history. You can also see this history by pressing `!` or `o`. See the `show_output` function in debugger.py.

asmeurer:
You wouldn't necessarily even need to show confirmation, although it might not hurt. It just makes it so that you can have the output visible once PuDB exits, and also not break the UI.

mvanderkamp:
I'm not sure I'm clear on how this is done yet- do I have access to that `screen` object from outside the DebuggerUI? If so, do you mean just call `screen.clear()` after each log line, or do the start/stop pairing? Or, do you mean pause the event loop (if I read that right, that's what happens on `n` and `c` keypresses, but I could be mistaken). I'll look into urwid a bit more perhaps, curses is generally a bit magical to me still.

mvanderkamp:
This should do the trick but if start/stop would be more efficient I can do that instead.

pudb/lowlevel.py
```diff
@@ -28,6 +28,39 @@
...
+            dbg.ui.screen.clear()
+
+
+ui_formatter = logging.Formatter(
+    fmt='\n*** Pudb UI Exception Encountered: %(message)s ***\n'
+)
+ui_handler = SafeStreamHandler()
+ui_handler.setFormatter(ui_formatter)
+ui_log = logging.getLogger('ui')
+ui_log.setLevel(logging.CRITICAL)
```
inducer, line 52: Given fd083e4, does it still make sense to swallow errors by default? (And, similarly, does the `--log-errors` option still make sense?)

mvanderkamp: I see your point! I've removed the option and set the errors to always be shown

pudb/lowlevel.py
```diff
@@ -28,6 +28,37 @@
...
+class SafeStreamHandler(logging.StreamHandler):
+    """
+    Logging stream handler that clears up the UI when it's done, if needed.
+    """
+    def emit(self, record):
+        from pudb import _have_debugger, _get_debugger
+        super(SafeStreamHandler, self).emit(record)
+        if _have_debugger():
+            dbg = _get_debugger()
+            dbg.ui.screen.clear()
```
inducer, line 43: Does this wait for the user to see the error message? If not, would it make sense to add that? Alternatively, given that we have a way of getting to the `DebuggerUI` object from here, we could simply call `DebuggerUI.message()` with the content of the message.

mvanderkamp: I'd rather not do that, since the log fires quite regularly when there's a problem in the variable view- once per refresh in fact, if memory serves, so requiring user input every time would render the session unusable. We'd have to figure out a way of caching prior messages and only logging them the first time, which could be possible I suppose, but might also make it harder to figure out why a message is appearing, since it won't reappear.

inducer: That's a fair point. As an alternative, we could direct those messages to pudb's built-in console (which we would have to force-show in the event of output to make sure the messages are seen by the user). How does that sound?

mvanderkamp: I could go with that. I'd want to still send them to stderr as well, or perhaps provide an option to send them to a file instead.

inducer: That sounds reasonable. Thanks for being receptive to the suggestion.

pudb/debugger.py
```diff
@@ -1595,6 +1595,7 @@ def add_cmdline_content(s, attr):
             self.cmdline_list.set_focus_valign("bottom")
             self.cmdline_list.set_focus(len(self.cmdline_contents) - 1,
                     coming_from="above")
+        self.add_cmdline_content = add_cmdline_content
```
inducer, line 1598: This seems like it wants to be a simple method.

pudb/debugger.py
```diff
@@ -2613,7 +2614,8 @@ def make_frame_ui(frame_lineno):
                 try:
                     class_name = frame.f_locals["self"].__class__.__name__
                 except Exception:
-                    pass
+                    from pudb.lowlevel import ui_log
+                    ui_log.exception('Failed to determine class name')
```
inducer, line 2618: `class_name` should perhaps be updated to reflect that an error occurred when trying to find it.

pudb/var_view.py
```diff
@@ -285,15 +287,15 @@ def type_stringifier(value):
...
+            ui_log.exception('string safe type stringifier failed')
 
     elif hasattr(type(value), "safely_stringify_for_pudb"):
         try:
             # (E.g.) Mock objects will pretend to have this
             # and return nonsense.
             result = value.safely_stringify_for_pudb()
         except Exception:
-            pass
+            ui_log.exception('safely_stringify_for_pudb call failed')
```
inducer, line 298: Perhaps update `result` to reflect a brief version of the error? (And point to the internal shell for traceback?)

pudb/var_view.py
```diff
@@ -285,15 +287,15 @@ def type_stringifier(value):
         try:
             return text_type(value)
         except Exception:
-            pass
+            ui_log.exception('string safe type stringifier failed')
```
inducer, line 290: Perhaps update `result` to reflect a brief version of the error? (And point to the internal shell for traceback?)

pudb/debugger.py
inducer, line 1598: I think it would be useful if `add_cmdline_content` could force-show the shell (at least until the next time it's focused in the case of auto-hiding) to ensure the output is discoverable.

inducer (commented):
Thanks for your work on this. I've tested it and I like it a lot. I've left some final comments. Good to merge after these are addressed from my end.

mvanderkamp:
Okay I _think_ that addresses your concerns. I wasn't quite sure about the approach for forcing the command line window, but I believe this does what you describe.

inducer:
This is lovely! Thanks so much for working on this.

PR #412, opened 2020-11-30
Make sure we can actually look up the values for keys to containers in the variable view

mvanderkamp:
Defining keys() but not __getitem__ is a pretty big code smell, but that
doesn't mean we should crash.

### DIFF
--- a/pudb/var_view.py
+++ b/pudb/var_view.py
@@ -417,7 +417,7 @@ class ValueWalker:
             if key_it is None:
                 try:
                     len_value = len(value)
-                except TypeError:
+                except (AttributeError, TypeError):
                     # no __len__ defined on the value, not worth mentioning!
                     pass
                 except Exception:
@@ -425,7 +425,7 @@ class ValueWalker:
                 else:
                     try:
                         value[0]
-                    except IndexError:
+                    except (LookupError, TypeError):
                         key_it = []
                     except Exception:
                         ui_log.exception("Item is not iterable")
@@ -445,7 +445,7 @@ class ValueWalker:
 
                     try:
                         next_value = value[key]
-                    except TypeError:
+                    except (LookupError, TypeError):
                         ui_log.exception("Failed to iterate an item that "
                             "appeared to be iterable.")
                         break
--- a/pudb/var_view.py
+++ b/pudb/var_view.py
@@ -432,7 +432,7 @@ class ValueWalker:
                     else:
                         key_it = xrange(len_value)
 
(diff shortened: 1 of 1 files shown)


pudb/var_view.py
```diff
@@ -432,7 +432,7 @@ def walk_value(self, parent, label, value, id_path=None, attr_prefix=None):
                     else:
                         key_it = xrange(len_value)
 
-            if key_it is not None:
+            if key_it is not None and hasattr(value, '__getitem__'):
```
inducer, line 435: Wouldn't it be more robust/Pythonic to handle failure of `value[key]`?

mvanderkamp: I'll take a look at this. I've also encountered some other similar instances where we could be a bit more robust about catching errors when trying to figure out how to walk a value.

mvanderkamp:
I think that should work. I was thinking of testing the `walk_value` method to make sure that all these various edge cases are handled and continue to behave reasonably, but I might do that separately unless you think that it would be a good thing to add here.

pudb/var_view.py
```diff
@@ -417,15 +417,15 @@ def walk_value(self, parent, label, value, id_path=None, attr_prefix=None):
             if key_it is None:
                 try:
                     len_value = len(value)
-                except TypeError:
+                except (AttributeError, TypeError):
```
inducer, line 420: Under what circumstances does `AttributeError` happen here?
```py
>>> class A:
...    pass
... 
>>> a = A()
>>> len(a)
  File "<stdin>", line 1, in <module>
TypeError: object of type 'A' has no len()
```

inducer (approved):
Thanks, LGTM! One question, but good to go at any rate.

inducer:
>  I was thinking of adding unit tests for the walk_value method to make sure that all these various edge cases are handled and continue to behave reasonably, but I might do that separately unless you think that it would be a good thing to add here.

I like this idea. Thanks for offering to add that.

pudb/var_view.py
mvanderkamp: You know, I'm not sure. I added that at some point during a marathon debugging session at work but forgot to copy down the traceback, if there even was one. I don't mind removing it.

mvanderkamp: (it's possible it was even some sort of internal bug in a `__len__` implementation)

inducer: Thanks for answering! I don't think it hurts.

PR #413, opened 2020-12-18
Support ipython kernel as a shell

xywei:
This PR adds support of `ipython-kernel` as a shell candidate. When called, it launches an `ipython` kernel, which can be connected to using `qtconsole` by
```
ipython qtconsole --existing
```

In `qtconsole`, plots stick around as a part of the scrollback history, which can be exported easily as HTML. This makes a bit better experience when inspecting multiple plots.

### DIFF
--- a/doc/shells.rst
+++ b/doc/shells.rst
@@ -54,10 +54,10 @@ PuDB supports the following external shells.
 - `IPython <https://ipython.org/>`_
    The `IPython` shell can also be used in a server-client fasion, which is
    enabled by selecting the shell `ipython_kernel` in the settings. When set,
-   the ``!`` key will start an `IPython` kernel and wait for connection from
-   `qtconsole`. Like other shells, `ipython_kernel` blocks the debugger UI
-   while it is active. Type `quit` or `exit` from a client to exit the kernel
-   and return to the debugger.
+   the ``!`` key will start an `IPython` kernel and wait for connection from,
+   e.g., `qtconsole`. Like other shells, `ipython_kernel` blocks the debugger
+   UI while it is active. Type `quit` or `exit` from a client to exit the
+   kernel and return to the debugger.
 - `bpython <https://bpython-interpreter.org/>`_
 - `ptpython <https://github.com/jonathanslenders/ptpython>`_
 
--- a/doc/shells.rst
+++ b/doc/shells.rst
@@ -52,15 +52,15 @@ PuDB supports the following external shells.
 - Internal (same as pressing ``Ctrl-x``). This is the default.
 - Classic (similar to the default ``python`` interactive shell)
 - `IPython <https://ipython.org/>`_
+   The `IPython` shell can also be used in a server-client fasion, which is
+   enabled by selecting the shell `ipython_kernel` in the settings. When set,
+   the ``!`` key will start an `IPython` kernel and wait for connection from
+   `qtconsole`. Like other shells, `ipython_kernel` blocks the debugger UI
+   while it is active. Type `quit` or `exit` from a client to exit the kernel
+   and return to the debugger.
 - `bpython <https://bpython-interpreter.org/>`_
 - `ptpython <https://github.com/jonathanslenders/ptpython>`_
 
-The `IPython` shell can also be used in a server-client fasion, which is
-enabled by selecting the shell `ipython_kernel` in the settings. When set, the
-``!`` key will start an `IPython` kernel and wait for connection from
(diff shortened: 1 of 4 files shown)


pudb/debugger.py
```diff
@@ -1920,6 +1920,8 @@ def run_external_cmdline(w, size, key):
             import pudb.shell as shell
             if CONFIG["shell"] == "ipython" and shell.have_ipython():
                 runner = shell.run_ipython_shell
+            elif CONFIG["shell"] == "ipython-kernel" and shell.have_ipython():
```
inducer, line 1923:
```suggestion
            elif CONFIG["shell"] == "ipython_kernel" and shell.have_ipython():
```

pudb/shell.py
```diff
@@ -231,6 +231,17 @@ def run_ipython_shell(globals, locals):
...
+
+def run_ipython_kernel(globals, locals):
+    from IPython import embed_kernel
+
+    class DummyMod(object):
+        pass
+
+    user_module = DummyMod()
+    user_module.__dict__ = globals
+    embed_kernel(module=user_module, local_ns=locals)
```
inducer, line 243: How do you make this function return? (Maybe add a note to the docs about this?

inducer, line 243: How does qtconsole handle reconnection? (i.e. kernel ends and then comes back?) (Maybe add a note to the docs about this?)

inducer (commented):
Interesting, thanks!

pudb/shell.py
xywei: `embed_kernel` blocks the execution until the kernel exits. Added docs to explain this.

xywei: Reconnection may be possible, but I am not sure. The qtconsole tab connecting to the kernel will be closed after sending "quit". But the kernels spawned later in the same debugging session seem to reuse the `kernel-xxxx.json` file.

doc/shells.rst
```diff
@@ -55,6 +55,13 @@ PuDB supports the following external shells.
 - `bpython <https://bpython-interpreter.org/>`_
 - `ptpython <https://github.com/jonathanslenders/ptpython>`_
 
+The `IPython` shell can also be used in a server-client fasion, which is
+enabled by selecting the shell `ipython_kernel` in the settings. When set, the
+``!`` key will start an `IPython` kernel and wait for connection from
```
inducer, line 60:
```suggestion
``!`` key will start an `IPython` kernel and wait for connection from, e.g.,
```

doc/shells.rst
```diff
@@ -55,6 +55,13 @@ PuDB supports the following external shells.
 - `bpython <https://bpython-interpreter.org/>`_
 - `ptpython <https://github.com/jonathanslenders/ptpython>`_
 
+The `IPython` shell can also be used in a server-client fasion, which is
```
inducer, line 58: Make this paragraph part of the enumeration above.

doc/shells.rst
```diff
@@ -55,6 +55,13 @@ PuDB supports the following external shells.
...
+enabled by selecting the shell `ipython_kernel` in the settings. When set, the
+``!`` key will start an `IPython` kernel and wait for connection from
+`qtconsole`. Like other shells, `ipython_kernel` blocks the debugger UI while
+it is active. Send "quit" from a client to exit the kernel and return to the
```
inducer, line 62: How does one send "quit" from the Qt console?

https://qtconsole.readthedocs.io/en/stable/ does not shed any light.

pudb/shell.py
inducer: OK, thx.

doc/shells.rst
xywei: It just means typing `quit<enter>` from a connected qtconsole. It is an IPython thing https://ipython.readthedocs.io/en/stable/api/generated/IPython.core.autocall.html#IPython.core.autocall.ExitAutocall. Will make the doc on this more clear.

doc/shells.rst
xywei: Fixed in 7258654

doc/shells.rst
xywei: Fixed in b4d73d3

doc/shells.rst
xywei: Changed the wording from "Send" to "Type" to avoid possible confusion in b4d73d3

inducer:
Thanks!

PR #417, opened 2021-01-01
Refactor var_view to use abstract base classes for detecting containers

mvanderkamp:
So: while testing I encountered various bugs related to how we were trying to determine whether something was a container that we could iterate. I thought about it for a bit, and the solution I came up with was to use some abstract base classes for detecting containers, and providing separate routines for each kind of container so that we can be sure that the class is sufficiently well-defined that we can actually inspect the container the way we want to.

I used custom ABCs instead of the built-in ones since the built-ins are actually quite restrictive and require that a lot of methods be present which we really don't need for our purposes.

Very much open to thoughts on this.


pudb/py3compat.py
```diff
@@ -19,6 +56,47 @@ def execfile(fname, globs, locs=None):
...
+        @classmethod
+        def __subclasshook__(cls, c):
+            if cls is PudbMapping:
+                return all([
+                    any("__getitem__" in b.__dict__ for b in c.__mro__),
+                    any("__iter__" in b.__dict__ for b in c.__mro__),
+                    any("keys" in b.__dict__ for b in c.__mro__),
+                ])
+            return NotImplemented
+    # }}}
```
inducer, line 99: We can simply drop Py2-specific bits at this point. (`setup.py` says pudb is 3.6+ only, and has for a while.) (And this should probably live in `var_view`.)

pudb/var_view.py
```diff
@@ -386,84 +470,25 @@ def walk_value(self, parent, label, value, id_path=None, attr_prefix=None):
...
-                if not cnt:
-                    self.add_item(new_parent_item, "<empty>", None)
+            if isinstance(value, PudbMapping):
+                self.walk_mapping(new_parent_item, label, value, id_path)
+                return
+            elif isinstance(value, PudbSequence):
+                self.walk_sequence(new_parent_item, label, value, id_path)
+                return
+            elif isinstance(value, PudbCollection):
+                self.walk_collection(new_parent_item, label, value, id_path)
```
inducer, line 481: This question is not specific, but it's potentially worth asking: Is there a reason to worry about types that are one of these *and* have interesting attributes?

pudb/var_view.py
```diff
@@ -347,6 +348,89 @@ class ValueWalker:
     def __init__(self, frame_var_info):
         self.frame_var_info = frame_var_info
 
+    def add_continuation_item(self, parent, id_path, count):
+        """
+        :param VariableWidget or None parent
+        :param str id_path:
+        :param int count:
+        :returns: True if a continuation item ("...") was added, else False
```
inducer, line 356: Just use py3 type annotations.

pudb/var_view.py
inducer, line 481: Would `singledispatch` work for this? (AKA how dependent are we on the order of these tests?)

pudb/var_view.py
inducer, line 356: Explain the logic better here. I'm guessing "a continuation item was not added" means that the calling walk should continue?

pudb/py3compat.py
```diff
@@ -11,6 +11,43 @@
...
+                ])
+            return NotImplemented
+
+    class PudbSequence(ABC):
+        @classmethod
+        def __subclasshook__(cls, c):
+            if cls is PudbSequence:
+                return all([
+                    any("__getitem__" in b.__dict__ for b in c.__mro__),
+                    any("__iter__" in b.__dict__ for b in c.__mro__),
```
inducer, line 35: Given that sequences end up with numerical indices, I'm worried that these tests aren't really specific enough. (E.g. `set` and `dict` would match.)

test/test_var_view.py
```diff
@@ -32,3 +41,309 @@ def test_get_stringifier():
...
+        "index",
+        "items",
+        "keys",
+        "values",
+    ])
+
+    # Deliberately starting from 0
+    for r in range(0, len(methods) + 1):
+        for selected_methods in sorted(
+                map(sorted, itertools.combinations(methods, r))):
```
inducer, line 110: Lol. :)

How many classes does this generate? How long do the resulting tests take?

test/test_var_view.py
```diff
@@ -32,3 +41,309 @@ def test_get_stringifier():
...
+        try:
+            return func(*args, **kwargs)
+        except KeyError:
+            # Classes without __iter__ are expected to raise IndexError in this
+            # sort of case. Frustrating, I know.
+            if (method_name == "__getitem__"
+                    and args and isinstance(args[0], integer_types)):
+                raise IndexError
+            raise
+    return method
```
inducer, line 141: Move above `containerlike_class_generator` for more linear reading?

test/test_var_view.py
```diff
@@ -32,3 +41,309 @@ def test_get_stringifier():
...
+class SetWithOverridenBool(set):
+    def __init__(self, iterable, truthy=True):
+        super(SetWithOverridenBool, self).__init__(iterable)
+        self.truthy = truthy
+
+    def __bool__(self):
+        return self.truthy
+
+
+def containerlike_class_generator():
```
inducer, line 92:
```suggestion
def generate_containerlike_class():
```

inducer (commented):
Thanks for working on this! I like the idea; it's a lot more systematic than what was there before. Some thoughts below.

mvanderkamp:
Thanks for the feedback! I'll patch this up!

pudb/py3compat.py
mvanderkamp: Right! Sorry I forgot- I'm unforunately stuck in a situation at work where I still have to use Python 2, I'll move this stuff to my branch that maintains py2 support.

pudb/var_view.py
mvanderkamp: Yeah I was wondering about this too actually. Even if we ditch the ABCs and use plain isinstance checks, this could still be a problem if someone, for example, subclasses `list` to add some interesting attributes.

Would it make sense to provide a "meta" attribute of some sort that displays contents, and expand the contents under that attribute?

mvanderkamp: With the way the ABCs are set up right now there's an ordering component: a mapping is a sequence which also has `keys` defined, and either one is likely to be recognized as a collection.

pudb/py3compat.py
mvanderkamp: Yup this is true: the way we're set up right now we're dependent on trying more specific types first.

test/test_var_view.py
mvanderkamp: 1. Lots :P
2. Not too long, at least for me (6 cores, 12 threads @ 3.2 GHz, 16GB RAM):
```
$ time pytest test/test_var_view.py
===================================================================================== test session starts =====================================================================================
platform linux -- Python 3.8.5, pytest-6.2.1, py-1.10.0, pluggy-0.13.1
rootdir: ---
collected 9 items

test/test_var_view.py .........

====================================================================================== 9 passed in 0.81s ======================================================================================

real    0m1.007s
user    0m0.959s
sys     0m0.048s
```

pudb/var_view.py
mvanderkamp: Actually another thought- this could apply to the "simple" types as well. Someone may well subclass `str` or `int` in some interesting way and want to inspect their added attributes.

test/test_var_view.py
mvanderkamp: More accurately: it generates 2048 classes.

pudb/var_view.py
inducer: I like the idea of the "meta" attribute. That might not even be all that yucky in terms of the traversal logic.

test/test_var_view.py
inducer: Oh OK. That's not too bad then. Thanks for checking!

pudb/var_view.py
mvanderkamp: Okay I've pushed an adjustment that allows expansion on all variables, regardless if they have a "basic" type, are a container, or are some other object, and uses a "metaitem" to hold the contents of containers.

The tests will now need fixing up of course, so this push should fail, but since the nature of the tests will change based on how this is implemented, I would prefer to wait until we've settled on a final approach before fixing them.

mvanderkamp: What do you think of the current approach to the "value string" for the contents meta-item? I might try putting together a "contents preview" since I'm not entirely happy with just showing "(hidden)".

mvanderkamp: I decided to go ahead and implement the preview.

mvanderkamp:
Noting here before I forget: this approach currently fails for, e.g., `pandas.Series`

mvanderkamp:
Okay there we go, I think that's properly ready now- sorry it took me a while to get the tests fixed up, and sorry I keep hitting lint errors. I was looking around for a ini or rc file to use for my linter but couldn't locate one- am I blind?

Anyway, curious to know what you think! I've been using this at work and it feels pretty good to me. It's really nice being able to expand the attributes of the basic types and containers.

mvanderkamp:
Another note to watch out for: the sequence walker breaks for 0d numpy arrays.

mvanderkamp:
I cleaned this up a bit to make better use of the container classes by moving iteration work out of the ValueWalker and into the container classes, which simplified the new ValueWalker code a fair bit. I think it also makes the new code easier to understand too.

There's also a fix for the scalar numpy array issue. All the containers catch TypeError when interating for expansion, but just ignore it for the preview to avoid being too noisy.

The other logical change is limiting the length of the preview string. Up to 3 items will be previewed, and each preview string will be up to 16 characters, so that expanding a container doesn't suddenly overwhelm the var_view if you have wrapping turned on.

inducer:
Resolved a merge conflict. Let me know if I did that right.

mvanderkamp:
Looks good to me!

mvanderkamp:
The "object X appears to be a <container> but doesn't behave like one" messages should use the container's label instead of the repr of the container so that the message isn't overwhelming and actually makes sense. Or perhaps the id path?

mvanderkamp:
I applied a couple minor fixes and merged master back in. This has been working well for me- is there anything else you'd like to see? How does the `<contents>` meta-item feel to you?

pudb/var_view.py
```diff
@@ -389,142 +611,101 @@ def walk_value(self, parent, label, value, id_path=None, attr_prefix=None):
...
-            except Exception:
-                ui_log.exception("Failed to obtain key iterator")
+        new_parent_item = self.add_item(parent, label, displayed_value,
+            id_path, attr_prefix)
+
+        if not iinfo.show_detail:
+            return
+
+        # containers --------------------------------------------------
+        if isinstance(value, (PudbCollection, PudbMapping, PudbSequence)):
```
inducer, line 632:
```suggestion
        if isinstance(value, CONTAINER_CLASSES):
```
and make `CONTAINER_CLASSES` a tuple?

pudb/var_view.py
```diff
@@ -389,142 +611,101 @@ def walk_value(self, parent, label, value, id_path=None, attr_prefix=None):
...
-                pass
-            except Exception:
-                ui_log.exception("Failed to obtain key iterator")
+        new_parent_item = self.add_item(parent, label, displayed_value,
+            id_path, attr_prefix)
+
+        if not iinfo.show_detail:
+            return
+
+        # containers --------------------------------------------------
```
inducer, line 631: I've come to like vim's fold-y comments, i.e.
```suggestion
        # {{{ show container contents
```
(with corresponding closing comment in the appropriate spot of course)

inducer (commented):
Thanks for working on this! The logic is much simpler and easier to follow this way, I really like it. Hope you don't mind my nitpicking below. Also, sorry about the long delay in reviewing. Life did a thing for a while.

inducer:
To cut down on noise in my email, I'm unsubscribing from this PR. When it next needs my attention, please @-mention me or hit the "request review" button. Otherwise, I may not see your messages in a timely manner.

pudb/var_view.py
```diff
@@ -364,21 +500,107 @@ def get_stringifier(iinfo):
...
+    BASIC_TYPES.append(type(None))
+    BASIC_TYPES.extend(integer_types)
+    BASIC_TYPES.extend(string_types)
+    BASIC_TYPES.extend((float, complex))
+    BASIC_TYPES = tuple(BASIC_TYPES)
+
+    NUM_PREVIEW_ITEMS = 3
+    MAX_PREVIEW_ITEM_LEN = 16
+
+    CONTENTS_LABEL = "<contents>"
```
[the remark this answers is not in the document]
mvanderkamp: You're absolutely right, I have noticed this myself- it often doubles the number of key presses before I see the thing I want to look at. I'll experiment with having the `<contents>` item start off expanded by default.

The downside to this approach is if a user is actually trying to look at the attributes of a container-like item. I suspect that's not the most common use case, and it's easy enough to collapse the `<contents>` item when you do want to do this. Another thing I might try implementing soon is to allow a "count" before a motion command, similar to vim, so that you can jump around more precisely (e.g. to skip over the contents to the attributes without having to collapse the contents, which I find myself wanting to do sometimes).

pudb/var_view.py
```diff
@@ -364,21 +500,107 @@ def get_stringifier(iinfo):
...
+        items = [cls._preview_entry(preview)
+                 for _, preview
+                 in zip(counter, container_cls.previews(container))]
+
+        preview = "{open}{items}{cont}{close}".format(
+            items=", ".join(items),
+            cont=", ..." if len(items) == cls.NUM_PREVIEW_ITEMS else "",
+            open=container_cls.SURROUNDS[0],
+            close=container_cls.SURROUNDS[1],
+        )
```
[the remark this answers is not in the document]
mvanderkamp: I can answer this and another comment in one go! ([this one](https://github.com/inducer/pudb/pull/417/files#r582396619))

The `previews` methods on the container abstract base classes truncate the individual preview items, so we don't have to worry about it here.

mvanderkamp:
That sounds more than fair! Thank you for taking the time to review my suggestions! I'll get right on with addressing your feedback and ping you once I think it's ready again!

pudb/var_view.py
```diff
@@ -52,6 +53,140 @@
...
+    @classmethod
+    def entries(cls, collection, label: str):
+        """
+        :yield: (label, entry, id_path_ext) tuples for each entry in the
+        collection.
+        """
+        assert isinstance(collection, cls)
+        try:
+            for count, entry in enumerate(collection):
+                yield None, entry, "[{k:d}]".format(k=count)
```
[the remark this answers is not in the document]
mvanderkamp: the `None` gets passed through to `VariableWidget`, which does two things:

1. There is logic through this class to convert a "falsy" `var_label` to the empty string as necessary.
2. The `render` method has special handling for when the `var_label` is `None`.

I find it a little bit confusing to be honest, but I didn't want to start refactoring that class as well in this PR.

pudb/var_view.py
```diff
@@ -389,142 +611,101 @@ def walk_value(self, parent, label, value, id_path=None, attr_prefix=None):
...
-                elif cnt_omitted_methods:
-                    label = "<omitted methods>"
-                else:
-                    label = "<empty>"
-                self.add_item(new_parent_item, label, None)
+        if not keys:
+            if cnt_omitted_private:
+                label = "<omitted private attributes>"
+            elif cnt_omitted_methods:
+                label = "<omitted methods>"
```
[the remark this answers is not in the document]
mvanderkamp: Looked into this. We're actually showing neither. This is only hit if `keys` is empty, in which case `cnt_omitted_private` and `cnt_omitted_methods` will never have a chance to be incremented.

I played around with tweaking this so that we do actually show the right thing, but I'm not sure what "the right thing" is in this case. I guess the UX value is explaining to users why they're not seeing anything under an expanded value and what to consider trying, but then what about the very common case of showing some, but not all of the things? We assume the user knows that they're not seeing everything and don't provide this hint. This is now inconsistent: we either give an indication of everything they're missing but only if they're not seeing anything (under an item that looks like maybe they could expand it to see those items, with no indication of how to actually start seeing those items), or trust that they know what they're doing.

I personally vote that we explicitly go the "trust the user to figure it out" route and ditch this bit of code. The "markers" already help point users in the right direction if they suspect they're not seeing everything.

pudb/var_view.py
mvanderkamp: Ooh yes those are snazzy! In this case though, I think what really needs to happen is for these long subsections to be extracted into their own methods.

pudb/var_view.py
mvanderkamp: Good idea, thanks!

pudb/var_view.py
```diff
@@ -52,6 +53,140 @@
...
+    @classmethod
+    def entries(cls, mapping, label: str):
+        """
+        :yield: (label, entry, id_path_ext) tuples for each entry in the
+        mapping.
+        """
+        assert isinstance(mapping, cls)
+        try:
+            for key in mapping.keys():
+                yield repr(key), mapping[key], "[{k!r}]".format(k=key)
```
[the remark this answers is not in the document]
mvanderkamp: Good catch!

pudb/var_view.py
```diff
@@ -364,21 +500,107 @@ def get_stringifier(iinfo):
...
     def __init__(self, frame_var_info):
         self.frame_var_info = frame_var_info
 
+    def add_continuation_item(self, parent: VariableWidget, id_path: str,
+                              count: int) -> bool:
+        """
+        :returns: True if a continuation item ("...") was added, else False.
+        If a continuation item was added, no further entries in the container
+        should be added. If no continuation item was added, continue adding
+        entries from the container.
```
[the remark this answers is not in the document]
mvanderkamp: Yup, will do!

pudb/var_view.py
```diff
@@ -52,6 +53,140 @@
...
+        assert isinstance(collection, cls)
+        try:
+            for count, entry in enumerate(collection):
+                yield None, entry, "[{k:d}]".format(k=count)
+        except (AttributeError, TypeError):
+            ui_log.error("Object {l!r} appears to be a collection, but does "
+                         "not behave like one.".format(l=label))
+
+    @classmethod
+    def previews(cls, collection):
```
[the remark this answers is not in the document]
mvanderkamp: This was for generating the "preview strings" for items in the container, to show beside the "<contents>" metaitem

mvanderkamp:
I've decided to try out an alternative approach:

Instead of using a "<contents>" meta-item, expand both contents and attributes directly under the expanded item, but similar to how we use the `".X"` syntax to mark an attribute `X`, use `"[X]"` syntax to mark a contained item with key/index `X`. For sets, I went with just empty brackets (`[]`).

This gets rid of the downgraded UX experience on regular lists and tuples, and should still be fine for custom containers, since we only display up to 10 items initially. It also, like the attributes, is reflective of python's syntax for accessing the item.

It also avoids all the str/repr issues with how to "preview" the contents.

pudb/var_view.py
```diff
@@ -52,6 +53,140 @@
...
+        except (AttributeError, TypeError):
+            ui_log.error("Object {l!r} appears to be a sequence, but does "
+                         "not behave like one.".format(l=label))
+
+    @classmethod
+    def previews(cls, sequence):
+        assert isinstance(sequence, cls)
+        try:
+            for entry in sequence:
+                yield str(entry)
```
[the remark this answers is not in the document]
mvanderkamp: Good point! I've removed the previews for the new approach.

pudb/var_view.py
```diff
@@ -52,6 +53,140 @@
...
+        except (AttributeError, TypeError):
+            ui_log.error("Object {l!r} appears to be a mapping, but does "
+                         "not behave like one.".format(l=label))
+
+    @classmethod
+    def previews(cls, mapping):
+        assert isinstance(mapping, cls)
+        try:
+            for key in mapping.keys():
+                yield "{k}: {v}".format(k=key, v=mapping[key])
```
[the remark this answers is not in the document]
mvanderkamp: Removed!

mvanderkamp:
@inducer Okay I _think_ I've addressed all your feedback. The big change now is that I've removed the `<contents>` meta-item since I agreed that it was a UX downgrade and have tried an alternative approach using `[]` annotation for contents to contrast with `.` annotation for attributes. I think I prefer this approach and I think it should feel comfortable to existing users since their experience will largely be the same, but now they also see the attributes for containers after the contents. As one way to highlight the advantages offered by this ability, I added a namedtuple to the `debug_me` script.

debug_me.py
```diff
@@ -1,3 +1,8 @@
+from collections import namedtuple
+
+Colour = namedtuple('Colour', ['red', 'green', 'blue', 'alpha'])
```
inducer, line 3:
```suggestion
Color = namedtuple('Color', ['red', 'green', 'blue', 'alpha'])
```
(personally I don't care, though the whole rest of the code base uses the AE spelling)

debug_me.py
```diff
@@ -19,6 +24,7 @@ def simple_func(x):
     y = dict((i, i**2) for i in s)
 
     k = set(range(5, 99))
+    c = Colour(137, 214, 56, 88)
```
inducer, line 27:
```suggestion
    c = Color(137, 214, 56, 88)
```

pudb/var_view.py
```diff
@@ -364,21 +478,112 @@ def get_stringifier(iinfo):
...
+    EMPTY_LABEL = "<empty>"
+    CONTINUATION_LABEL = "[...]"
+
     def __init__(self, frame_var_info):
         self.frame_var_info = frame_var_info
 
+    def add_continuation_item(self, parent: VariableWidget, id_path: str,
+                              count: int, length: int) -> bool:
+        """
+        :returns: True if a continuation item ("[...]") was added, else False.
```
inducer, line 500:
```suggestion
        :arg length: the total length of the container. Negative if not known.
        :returns: True if a continuation item ("[...]") was added, else False.
```

pudb/var_view.py
inducer: :+1:

inducer (approved):
This is super nice! Thank you for working on it! I'll apply my suggestions and set it to automerge once CI passes.

PR #418, opened 2021-01-05
Catch and log unexpected exceptions emenating from the event loop

mvanderkamp:


pudb/debugger.py
```diff
@@ -2504,7 +2504,10 @@ def event_loop(self, toplevel=None):
                     if k == "window resize":
                         self.size = self.screen.get_cols_rows()
                     else:
-                        toplevel.keypress(self.size, k)
+                        try:
+                            toplevel.keypress(self.size, k)
+                        except Exception:
+                            ui_log.exception("Encountered unexpected exception in the debugger UI.")
```
inducer, line 2510: Maybe add some text telling people to look in the console window for the exception text?

inducer (commented):
Thanks! LGTM for the most part. Flake8 complained, and I had a thought on discoverability.

pudb/debugger.py
mvanderkamp: Where would I put the text? In a popup?

inducer: Ah, nvm. This message itself isn't very visible. Fair, we can just bury it in the UI log for now. It's not likely to be very interesting to end users anyway.

inducer: Hmm, OTOH. I think this merits a popup, because likely that means whatever the user asked us to do didn't happen.

#419

Issue #419, opened 2021-01-06
Show a popup for exceptions caught in the event loop


inducer:
Follow-up from https://github.com/inducer/pudb/pull/418#discussion_r552771667

cc @mvanderkamp

mvanderkamp:
This can be closed now, right?

inducer:
Yep, thanks!

PR #420, opened 2021-01-08
Use a popup to display internal errors

mvanderkamp:
For issue #419 
I cleaned up the existing code for showing uncaught exceptions so that I could reuse it, and while testing I found another possible place where it might be beneficial to catch internal exceptions.


inducer:
LGTM! Thanks for working on this.

PR #422, opened 2021-01-18
Add a theme style for current line column

mvanderkamp:
I like to use a different background for the line number column, then again for the sign column in vim. It was almost possible to replicate this in pudb before, the only thing missing was that the column for the ">" current line marker didn't have its own style setting available. Now it does!

I made sure to go back and include it in all the previously existing themes, and I quickly tried them out and they seemed to work. What do you think of the name of the style, is it clear enough?

### DIFF
--- a/pudb/source_view.py
+++ b/pudb/source_view.py
@@ -105,7 +105,7 @@ class SourceLine(urwid.FlowWidget):
             line_prefix = self.line_nr
 
         line_prefix = crnt+bp+line_prefix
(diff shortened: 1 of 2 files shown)


inducer:
I have no problem with the idea of this change. Is `current line marker` too long?

mvanderkamp:
Nope, thanks for the suggestion! Done!

inducer:
Thx!

PR #424, opened 2021-01-21
Clean up some render issues with long vars/values in the var_view

mvanderkamp:
This PR just deals with some pretty minor things in the way the var_view renders variables with long names and/or values.

If you have a really long variable/value combo in the var view and you don't wrap, the value is printed on a subsequent line, e.g.
```
| | .my_long_named_attribute
| |   [0, 1, 2, 3, 4, 5]
```
If you were to wrap the same variable it might be rendered as:
```
| | .my_long_named_attribute: [0, 1,
  | | 2, 3, 4, 5]
```
---
This PR normalizes the two by making the non-wrapped one look like this: (note the ":" on the first line)
```
| | .my_long_named_attribute:
| |   [0, 1, 2, 3, 4, 5]
```
And the wrapped one look like this: (note the spaces on the second line now come after the prefix)
```
| | .my_long_named_attribute: [0, 1,
| |   2, 3, 4, 5]
```
There's also a tweak to make sure that when not wrapping, we'll only use two lines for a variable if we actually don't have enough space to show it in one.

### DIFF
--- a/pudb/var_view.py
+++ b/pudb/var_view.py
@@ -145,9 +145,11 @@ class VariableWidget(urwid.FlowWidget):
     def selectable(self):
         return True
 
-    SIZE_LIMIT = 20
-
-    def _get_text(self, size):
+    def _get_wrapped_lines(self, size):
+        """
+        :return: list of string lines, including prefixes, wrapped to fit in
+        the available space,
+        """
         maxcol = size[0] - len(self.prefix)  # self.prefix is a padding
         var_label = self.var_label or ""
         value_str = self.value_str or ""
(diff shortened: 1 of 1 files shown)


pudb/var_view.py
```diff
@@ -145,9 +145,11 @@ def __str__(self):
     def selectable(self):
         return True
 
-    SIZE_LIMIT = 20
-
-    def _get_text(self, size):
+    def _get_wrapped_lines(self, size):
+        """
+        :return: list of string lines, including prefixes, wrapped to fit in
+        the available space,
```
inducer, line 151: Document `size`?

pudb/var_view.py
```diff
@@ -145,9 +145,11 @@ def __str__(self):
...
+        """
         maxcol = size[0] - len(self.prefix)  # self.prefix is a padding
```
inducer, line 153:
```suggestion
        maxcol = size[0] - len(self.prefix)  # self.prefix is padding
```

inducer (approved):
Thanks! Looks good to me except for these wrinkles.

mvanderkamp:
I did a bit of refactoring to unpack the `size` tuple earlier, which seems more clear to me, and added documentation.

inducer:
LGTM, thanks!

PR #425, opened 2021-01-21
Do not accidentally repeat similar variable names at top

mvanderkamp:
This fixes a bug where if you have a variable 'foo' and another variable 'foobar', and use the '@' command to repeat 'foo' at the top of the variable list, 'foobar' would also get repeated.


inducer:
LGTM. Thanks!

Issue #401, opened 2020-08-21
Yucky console error message when custom stringifier not found


endlisnis:
Sometimes, when I'm using pudb to step through code and inspect variables, I will get this message popping up:

```
Error when importing custom stringifier:
Traceback (most recent call last):
  File "/home/rolf/.local/lib/python3.8/site-packages/pudb/var_view.py", line 315, in get_stringifier
    execfile(expanduser(iinfo.display_type), custom_stringifier_dict)
  File "/home/rolf/.local/lib/python3.8/site-packages/pudb/py3compat.py", line 13, in execfile
    exec(compile(open(fname).read(), fname, 'exec'), globs, locs or globs)
FileNotFoundError: [Errno 2] No such file or directory: ''
Hit enter:
```

I'm not using a customer stringifier that I'm aware of, and hitting enter just gets me into a loop of this error printing out.

I'm using python3.8 on Linux Mint, pudb version "2019.2".

inducer:
Closed by #429.

inducer:
(Thanks @mvanderkamp!)

endlisnis:
Great.  When will this get pushed out to pip?  I'd like to give it a try.

Issue #384, opened 2020-04-29
pudb_stringifier(obj) not called for int()


jlescher:
I would like to be able to display ints as hexadecimal.
However it seems that pudb_stringifier(obj) is not called for int values.

Is there any reason why ?

asmeurer:
I guess the original idea is that the stringifier is only needed for unsafe types. But it would probably be better to implement a default stringifier and have the custom one override it (the default one can be part of the API so that custom ones can fall back to it).

inducer:
> But it would probably be better to implement a default stringifier and have the custom one override it (the default one can be part of the API so that custom ones can fall back to it).

I like this idea.

mvanderkamp:
I agree this would be useful. The changes in #417 would make this easier since it simplifies the code to test for all the basic types at once, but this is generally independent of that changeset so there's no need to wait on it.

Possibilities that I see:
1. Always use `repr()` by default for the basic types, but allow custom stringifiers
2. Don't treat the basic types as distinct from other types.
3. Provide a config option for toggling between 1. and 2.

inducer:
How about we simply move the check for the basic types into the default stringifier? (i.e. not "type", but "type-or-basic-value")

asmeurer:
Isn't that what is already happening at https://github.com/inducer/pudb/blob/da44ac39d956cf0a73cfc3b5969f4b86f500a1c4/pudb/var_view.py#L289? The only thing I don't like about that is that the name "type" is misleading. IMO "type" should literally just be `type` and the current `type_stringifier` should be called something else.

mvanderkamp:
Interesting, I think I like that idea. How about "default" for the existing one?

inducer:
"default" is fine by me. Essentially, [this bit of code](https://github.com/inducer/pudb/blob/da44ac39d956cf0a73cfc3b5969f4b86f500a1c4/pudb/var_view.py#L376-L381) would simply have to go, to give more power to the stringifiers.

asmeurer:
I like "safe" for a name. I think that's what we already call it in some places anyway. Although "default" is better for a keyboard shortcut (s is already taken for str).

inducer:
#384 should be addressed via its number-anagram #438 by @mvanderkamp. :)

PR #429, opened 2021-02-01
Add a built-in stringifier to display the object's id()

mvanderkamp:
I've been using this as a custom stringifier for a while, but I think it is useful and generic enough to warrant inclusion as a built-in.

Use cases typically involve quickly checking whether or not two "equal" variables point to the same object.

Thoughts? I think I hooked this up in all the right places. I've also thrown in a change to use the new `ui_log` to print errors when trying to run the custom stringifier, since currently the message mucks with the rendering of the debugger.

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -112,7 +112,7 @@ Keys in variables list:
     \/enter/space - expand/collapse
     h - collapse
     l - expand
-    t/r/s/c - show type/repr/str/custom for this variable
+    t/r/s/i/c - show type/repr/str/id/custom for this variable
     H - toggle highlighting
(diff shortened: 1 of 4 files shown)


inducer:
Thanks! When I played around with it, I was a bit surprised that it didn't work on `int` and `str` (among a few other types), but that's not the fault of this PR. (Do you think that's worth addressing?)

At any rate, LGTM! Thanks for your contribution.

mvanderkamp:
Yeah I do think that's worth addressing. I'm just looking through the issues list...

- I think this closes #401 
- Looks like there is an issue for addressing this: #384

Issue #211, opened 2016-12-31
Question: is it possible to open file in $EDITOR at the current cursor line?


alok:
I often find a bug and then have to open up an editor to, well, *edit* it. It'd be nice if I could just hit a key.

cdman:
Hmm, a quick search seems to indicate that both `vim` (`vi`?) and `nano` support the `$EDITOR +$LINE_NUMBER $FILE_NAME` syntax, so this could be made to work for at least a large amount of cases. However I worry that it might be confusing since Python doesn't provide auto-reload out of the box... (so you debug your file, edit it, close the editor and then you're back at the unchanged file in the debugger?).

asmeurer:
After editing it could pop up a dialog asking to continue with the pre-edited version or to quit. Maybe also restart with `reload` with the caveat that `reload` can sometimes introduce its own bugs.

huangyuyao:
Any update?

asmeurer:
No one is working on it as far as I know. It shouldn't be too hard to implement.

gaborvecsei:
I'd also love this feature!

gaborvecsei:
I created a pull request which solves this issue: #430

gaborvecsei:
My PR is merged. From now on, you can enjoy file editing 😊 Just hit `Ctrl-e`. 😉 

(I think this issue can be closed)

bostone:
This doesn't work if `nano` editor is missing which is my case (I'm in Docker container with only vi available)

asmeurer:
You should be able to fix it by setting `EDITOR=vi` in the docker environment. `nano` is used as a default if `EDITOR` is not set.

inducer:
Closing as per discussion above, cf. #430.

PR #430, opened 2021-02-26
Open file to edit from debugger

gaborvecsei:
Issue: #211 

## Feature

When you are on a file at the "stack" window, you can press `Ctrl+e` and the file will open with your editor, based on the `$EDITOR` environment variable. When you are done with editing you can just save and quit the file, and you will be returned to the debugger.

`reload()` is not used (because of this comment: https://github.com/inducer/pudb/issues/211#issuecomment-327031721), so the edits won't be noticable immediately, only after quitting the debugger and starting it again

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -1145,7 +1145,6 @@ class DebuggerUI(FrameVarInfoKeeper):
             file_name = curframe.f_code.co_filename
 
             open_file_editor(file_name, line_number)
-            redraw_screen(w, size, key)
 
         self.stack_list.listen("ctrl e", open_editor_on_stack_frame)
 
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -346,16 +346,11 @@ class Debugger(bdb.Bdb):
         self.curframe, line_number = self.stack[index]
         filename = self.curframe.f_code.co_filename
 
+        if not line_number:
+            line_number = 1
+
         editor = os.environ.get("EDITOR", "nano")
-        editor = editor.lower()
-
-        editor_command = f"{editor}"
-        if editor in {"vi", "vim", "nvim", "emacs"}:
-            editor_command = f"{editor} +{line_number}"
-        elif editor == "nano":
-            editor_command = f"{editor} -l {line_number}"
-
-        command_to_run = f"{editor_command} {filename}"
+        command_to_run = f"{editor} +{line_number} {filename}"
         os.system(command=command_to_run)
 
     def move_up_frame(self):
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -1133,8 +1133,7 @@ class DebuggerUI(FrameVarInfoKeeper):
             except Exception:
                 from pudb.lowlevel import format_exception
                 self.message("Exception happened when trying to edit the file:"
-                             "\n\n%s" % (
-                    "".join(format_exception(sys.exc_info()))),
+                             "\n\n%s" % ("".join(format_exception(sys.exc_info()))),
                     title="File Edit Error")
                 return
 
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -344,7 +344,8 @@ class Debugger(bdb.Bdb):
 
         self.ui.stack_list._w.set_focus(self.ui.translate_ui_stack_index(index))
 
-    def open_file_to_edit(self, filename, line_number):
+    @staticmethod
+    def open_file_to_edit(filename, line_number):
 
         if not filename:
             raise ValueError("Could not get filename as it was None")
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -360,6 +360,8 @@ class Debugger(bdb.Bdb):
         import subprocess
         subprocess.call([editor, f"+{line_number}", filename], shell=False)
 
+        return filename
+
     def move_up_frame(self):
         if self.curindex > 0:
             self.set_frame_index(self.curindex-1)
@@ -1126,13 +1128,18 @@ class DebuggerUI(FrameVarInfoKeeper):
             index = self.translate_ui_stack_index(pos)
 
             try:
-                self.debugger.open_file_to_edit(index)
+                filename_edited = self.debugger.open_file_to_edit(index)
             except Exception:
                 from pudb.lowlevel import format_exception
                 self.message("Exception happened when trying to edit the file: \n\n%s" % (
                     "".join(format_exception(sys.exc_info()))),
                     title="File Edit Error")
+                return
 
+            self.message("File is changed, but the execution is continued with the 'old' codebase.\n"
+                         f"Changed file: {filename_edited}\n"
+                         "Please quit and restart to see changes",
+                         title="File is changed")
             redraw_screen(w, size, key)
 
         self.stack_list.listen("ctrl e", open_editor_on_frame)
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -1121,9 +1121,16 @@ class DebuggerUI(FrameVarInfoKeeper):
         self.stack_list.listen("enter", examine_frame)
 
         def open_file_editor(file_name, line_number):
+            file_changed = False
+
             try:
+                original_modification_time = os.path.getmtime(file_name)
+                self.screen.stop()
                 filename_edited = self.debugger.open_file_to_edit(file_name,
                                                                   line_number)
+                self.screen.start()
+                new_modification_time = os.path.getmtime(file_name)
+                file_changed = new_modification_time - original_modification_time > 0
             except Exception:
                 from pudb.lowlevel import format_exception
                 self.message("Exception happened when trying to edit the file:"
@@ -1131,11 +1138,12 @@ class DebuggerUI(FrameVarInfoKeeper):
                     title="File Edit Error")
                 return
 
-            self.message("File is changed, but the execution is continued with"
-                         " the 'old' codebase.\n"
-                         f"Changed file: {filename_edited}\n\n"
-                         "Please quit and restart to see changes",
-                         title="File is changed")
+            if file_changed:
+                self.message("File is changed, but the execution is continued with"
+                             " the 'old' codebase.\n"
+                             f"Changed file: {filename_edited}\n\n"
+                             "Please quit and restart to see changes",
+                             title="File is changed")
 
         def open_editor_on_stack_frame(w, size, key):
             _, pos = self.stack_list._w.get_focus()
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -123,6 +123,7 @@ Keys in variables list:
 
 Keys in stack list:
     enter - jump to frame
+    Ctrl-e - open file at line to edit with $EDITOR
 
 Keys in breakpoints list:
     enter - jump to breakpoint
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -2049,7 +2049,6 @@ class DebuggerUI(FrameVarInfoKeeper):
             file_name = self.debugger.curframe.f_code.co_filename
             line_number = self.debugger.curframe.f_lineno
             open_file_editor(file_name, line_number)
-            redraw_screen(w, size, key)
 
         self.top.listen("o", show_output)
         self.top.listen("ctrl r", reload_breakpoints)
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -342,6 +342,22 @@ class Debugger(bdb.Bdb):
 
         self.ui.stack_list._w.set_focus(self.ui.translate_ui_stack_index(index))
 
+    def open_file_to_edit(self, index):
+        self.curframe, line_number = self.stack[index]
+        filename = self.curframe.f_code.co_filename
+
+        editor = os.environ.get("EDITOR", "nano")
+        editor = editor.lower()
+
+        editor_command = f"{editor}"
+        if editor in {"vi", "vim", "nvim"}:
+            editor_command = f"{editor} +{line_number}"
+        elif editor == "nano":
+            editor_command = f"{editor} -l {line_number}"
+
+        command_to_run = f"{editor_command} {filename}"
+        os.system(command=command_to_run)
+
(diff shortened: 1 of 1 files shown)


pudb/debugger.py
```diff
@@ -342,6 +342,22 @@ def set_frame_index(self, index):
...
 
+    def open_file_to_edit(self, index):
+        self.curframe, line_number = self.stack[index]
+        filename = self.curframe.f_code.co_filename
+
+        editor = os.environ.get("EDITOR", "nano")
+        editor = editor.lower()
+
+        editor_command = f"{editor}"
+        if editor in {"vi", "vim", "nvim"}:
```
asmeurer, line 353: I believe emacs supports this as well.

asmeurer:
This needs to be added to the help. Maybe also with a note that changes will only take effect when pudb is restarted.

pudb/debugger.py
gaborvecsei: Good catch I'll add it!

gaborvecsei: Done

gaborvecsei:
> This needs to be added to the help. Maybe also with a note that changes will only take effect when pudb is restarted.

@asmeurer I'll add it. The question is where should I put this? Under `Features` in the README?

asmeurer:
You can add it to the docs if you want, but I actually meant the interactive help that currently lists every keyboard command (what pops when you press `?`).

pudb/debugger.py
```diff
@@ -342,6 +342,22 @@ def set_frame_index(self, index):
...
+        filename = self.curframe.f_code.co_filename
+
+        editor = os.environ.get("EDITOR", "nano")
+        editor = editor.lower()
+
+        editor_command = f"{editor}"
+        if editor in {"vi", "vim", "nvim", "emacs"}:
+            editor_command = f"{editor} +{line_number}"
+        elif editor == "nano":
+            editor_command = f"{editor} -l {line_number}"
```
asmeurer, line 356: According to `man nano` at least on my machine, nano also supports the `+line_number` syntax. Maybe we should just always do this regardless of the editor. I suspect other editors like vscode will probably also support this.

inducer, line 356: This fails if an explicit path is specified, or if the command is `emacs-gtk` or some-such. Maybe check for substrings? It might also be useful to let the user customize the invocation in settings.

pudb/debugger.py
```diff
@@ -342,6 +342,22 @@ def set_frame_index(self, index):
...
         self.ui.stack_list._w.set_focus(self.ui.translate_ui_stack_index(index))
 
+    def open_file_to_edit(self, index):
+        self.curframe, line_number = self.stack[index]
+        filename = self.curframe.f_code.co_filename
+
+        editor = os.environ.get("EDITOR", "nano")
+        editor = editor.lower()
+
+        editor_command = f"{editor}"
```
inducer, line 352: Could you move this to the `else` branch?

pudb/debugger.py
```diff
@@ -342,6 +342,22 @@ def set_frame_index(self, index):
 
         self.ui.stack_list._w.set_focus(self.ui.translate_ui_stack_index(index))
 
+    def open_file_to_edit(self, index):
+        self.curframe, line_number = self.stack[index]
+        filename = self.curframe.f_code.co_filename
```
inducer, line 347: AFAIK, this may be None. Show an appropriate message if that's the case, instead of crashing.

pudb/debugger.py
```diff
@@ -342,6 +342,22 @@ def set_frame_index(self, index):
...
+        editor = editor.lower()
+
+        editor_command = f"{editor}"
+        if editor in {"vi", "vim", "nvim", "emacs"}:
+            editor_command = f"{editor} +{line_number}"
+        elif editor == "nano":
+            editor_command = f"{editor} -l {line_number}"
+
+        command_to_run = f"{editor_command} {filename}"
+        os.system(command=command_to_run)
```
inducer, line 359: This will fail if the filename contains spaces. Use `subprocess` and don't invoke via the shell.

pudb/debugger.py
```diff
@@ -342,6 +342,22 @@ def set_frame_index(self, index):
 
         self.ui.stack_list._w.set_focus(self.ui.translate_ui_stack_index(index))
 
+    def open_file_to_edit(self, index):
+        self.curframe, line_number = self.stack[index]
+        filename = self.curframe.f_code.co_filename
+
+        editor = os.environ.get("EDITOR", "nano")
+        editor = editor.lower()
```
inducer, line 350: I'd remove this. Unix file systems can be case-sensitive, and somebody might have an upper-case letter in their `$EDITOR`.

pudb/debugger.py
inducer, line 347: It may also not be a valid filename. Maybe check before proceeding?

pudb/debugger.py
inducer, line 359: What completion semantics do we expect here? Should the source be reloaded when we come back from the editor? (This involves two possible parts: `linecache` and Python module reload) (If we do nothing, it'll be stale, which will be confusing.)

inducer (commented):
Thanks for working on this. A few thoughts below.

inducer:
To cut down on noise in my email, I'm unsubscribing from this PR. When it next needs my attention, please @-mention me or hit the "request review" button. Otherwise, I may not see your messages in a timely manner.

pudb/debugger.py
asmeurer: Yeah, I'm unsure about this too. Doing a `reload` is probably a bad idea, as it is wrong even more subtle ways (if you really want to do that, you can do it yourself manually in the console). Maybe after editing it should popup a message asking if you want to stop the debugger so you can restart the interpreter?

pudb/debugger.py
gaborvecsei: DONE

pudb/debugger.py
gaborvecsei: Because of a request above this is not needed.

Every editor supports the `+l flag` so we can just `f"{editor} +l {line_number} {filename}"`

pudb/debugger.py
gaborvecsei: Done

pudb/debugger.py
gaborvecsei: I think not using `reload` is the better option. I will display a message, so the user is aware that the file changed, and they can stop the execution and restart if they want

gaborvecsei: DONE

gaborvecsei:
Feature is added to the help (when pressing `?`)

gaborvecsei:
Hi @inducer, @asmeurer!
Thank you for the look at the PR.
I am done with all the requests, can you review the PR?

pudb/debugger.py
```diff
@@ -123,6 +123,7 @@
 
 Keys in stack list:
     enter - jump to frame
+    Ctrl-e - open file at line to edit with $EDITOR
```
asmeurer, line 126: Oh I didn't notice that this is based on the stack list. I was expecting this to be implemented in the main code view.

gaborvecsei: I understand why you think it would make more sense. I also thought about this and I implemented it here because of the resons:
- Maybe the developer would like to "look around" in the stack list for other files which are "effected" in the debugging. And it is easier with directly opening the files from stack list
- The `Ctrl-e` shortcut is useable/not allocated
- No (new) users can accidentaly open the file editor window. It takes 1 more shortcut to go to the stack list
- It is not confusing (even with the warning message) that the file is not changed which is in front of them after the edit.
- If we would open the current file which line would we jump to? The one under our cursor, the top/bottom/middle breakpoint? --> this can make some consusion, while on stack list the line number is displayed so no surprises will happen for the user

(maybe I am biased by my experience 😄 )

inducer: FWIW, I also didn't catch that this was on the stack list. So there's at least two data points on where folks would expect this functionality to live. (I'm not saying remove it from the stack, but also add it to the main source window, too.) As for what line to show, I think the current highlight is the obvious choice.

pudb/debugger.py
inducer: You could *offer* a reload in the message. Do you think that's reasonable?

pudb/debugger.py
gaborvecsei: I fixed it, now both on Main view and Stack view the `Ctrl-e` edit functionality is working
@inducer Can you review it?

pudb/debugger.py
```diff
@@ -1103,6 +1123,35 @@ def examine_frame(w, size, key):
...
+                self.message("Exception happened when trying to edit the file:"
+                             "\n\n%s" % ("".join(format_exception(sys.exc_info()))),
+                    title="File Edit Error")
+                return
+
+            self.message("File is changed, but the execution is continued with"
+                         " the 'old' codebase.\n"
+                         f"Changed file: {filename_edited}\n\n"
+                         "Please quit and restart to see changes",
+                         title="File is changed")
```
inducer, line 1141: This can claim that the code was changed, when it didn't, which might spook users. Maybe check the mtime with `stat`?

pudb/debugger.py
```diff
@@ -342,6 +344,24 @@ def set_frame_index(self, index):
 
         self.ui.stack_list._w.set_focus(self.ui.translate_ui_stack_index(index))
 
+    def open_file_to_edit(self, filename, line_number):
```
inducer, line 347: No real reason for this to be a method. It doesn't use `self` AFAICT. Move to a function?

pudb/debugger.py
```diff
@@ -342,6 +344,24 @@ def set_frame_index(self, index):
...
+
+        if not filename:
+            raise ValueError("Could not get filename as it was None")
+
+        if not os.path.isfile(filename):
+            raise FileNotFoundError(f"Could not find the file with name: {filename}")
```
inducer, line 353:
```suggestion
            raise FileNotFoundError(f"'{filename}' not found or is not a file.")
```

pudb/debugger.py
```diff
@@ -342,6 +344,24 @@ def set_frame_index(self, index):
 
         self.ui.stack_list._w.set_focus(self.ui.translate_ui_stack_index(index))
 
+    def open_file_to_edit(self, filename, line_number):
+
+        if not filename:
+            raise ValueError("Could not get filename as it was None")
```
inducer, line 350: Make this more actionable for a user, explaining what may have happened.

pudb/debugger.py
```diff
@@ -1103,6 +1123,35 @@ def examine_frame(w, size, key):
...
+
+        def open_editor_on_stack_frame(w, size, key):
+            _, pos = self.stack_list._w.get_focus()
+            index = self.translate_ui_stack_index(pos)
+
+            curframe, line_number = self.debugger.stack[index]
+            file_name = curframe.f_code.co_filename
+
+            open_file_editor(file_name, line_number)
+            redraw_screen(w, size, key)
```
inducer, line 1151: Why is this necessary? (Instead, you should stop/start the Urwid screen, which may switch the terminal between front and alternate buffers.)

inducer (commented):
Thanks for following up. Some more thoughts below.

pudb/debugger.py
gaborvecsei: And what would be the threshold where we can be certain there was a change? 1minute,  5 minutes?
Even if it is 30 minutes, sometimes I modify code for >30mins, even drink coffee, etc.

I just saying we cannot expect the users to quickly modify code when they are debugging.

What if the message would say something like

> You were in edit mode. If anything was changed in the code, please restart the debugging session, if not, you can continue in this session.

pudb/debugger.py
gaborvecsei: DONE

pudb/debugger.py
gaborvecsei: Actually as I see this can not happen.
I removed the raising of the exception

gaborvecsei: DONE

pudb/debugger.py
gaborvecsei: We don't need it 🙂  Good catch!

gaborvecsei: DONE

gaborvecsei:
@inducer I am done with the requested fixes, and for the "changed file popup" I left an answer.

pudb/debugger.py
inducer: I'm not sure I understand. If the mtime is different than before we started the editor, it's pretty clear that the file was changed.

pudb/debugger.py
```diff
@@ -2000,6 +2045,12 @@ def redraw_screen(w, size, key):
         def help(pages):
             self.message(pages, title="PuDB - The Python Urwid Debugger")
 
+        def edit_current_frame(w, size, key):
+            file_name = self.debugger.curframe.f_code.co_filename
+            line_number = self.debugger.curframe.f_lineno
+            open_file_editor(file_name, line_number)
+            redraw_screen(w, size, key)
```
inducer, line 2052: Elsewhere, I thought we got rid of `redraw_screen`?

gaborvecsei: Sorry I missed this, and removed only 1 from another place.
DONE

pudb/debugger.py
gaborvecsei: Sorry, I overthinked this one. Now it is implemented as requested.

DONE

pudb/debugger.py
gaborvecsei: (This conversation is outdated)

gaborvecsei:
@inducer Requests are fixed, please review

inducer:
This looks good to me. As you can see, I changed the behavior for Ctrl-e in the source window to use the focused rather than the current line. I hope you're OK with that. Thanks for your contribution!

gaborvecsei:
> This looks good to me. As you can see, I changed the behavior for Ctrl-e in the source window to use the focused rather than the current line. I hope you're OK with that. Thanks for your contribution!

Of course, I am okay with that. Thank you very much for the reviews!

Issue #414, opened 2020-12-23
Provide a simple way to install pudb's tab-completion


rbarrois:
Before #350, tab completion was provided natively through Python's readline module.
New versions rely on a third party dependency, which adds an extra step to setup ­— and requires user to install an extra package without a relation with pudb.

It would be great if users could directly install pudb and its tab-completion support in a single `pip install`:
- Either add a fallback on Python's `readline` module if `jedi` can't be found;
- Or add an optional extra that pulls in a compatible version of `jedi`, e.g `pip install pudb[completion]`

If one of those options interests you, I'll be glad to provide a pull-request accordingly ;)

And thanks again for pudb, it's an amazing program 😉

inducer:
The optional extra sounds good to me. But I'm also open to making jedi a hard dependency. What do you think? @asmeurer any thoughts?

rbarrois:
Just a quick note: on my setup, the first tab completion (with jedi) seems to take at least half a second, whereas it was nearly instant with readline :/

inducer:
If you'd like `rlcompleter` back as a fallback, I'm open to looking at a PR.

asmeurer:
The only issue is that rlcompleter causes issues on the system Python on macOS (see the issues corresponding to that PR). Simply importing it imports `readline`, which seems to cause issues. A solution could be to extract the dircompleter stuff from the rlcompleter module so that we can use it without importing readline. Or figure out why libedit breaks urwid and fix the bug (but good luck with that). I also wonder if Jedi has a faster dir() only completer option.

inducer:
@asmeurer Ah, thanks for for reminding me. In that case, no, we probably shouldn't bring `rlcompleter` back.

PR #432, opened 2021-03-04
Add jedi dependency

hexagonrecursion:
Fixes https://github.com/inducer/pudb/issues/414

pudb needs jedi for tab completion, but it was not listed as a
dependency. This PR adds the missing dependency.

I'm not sure which version of jedi to depend on. Is there any reason
not to depend on the latest version (currently 0.18)?

I have added the <1 constraint to guard against potential future
breaking changes in jedi

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,6 +28,7 @@ setup(
     install_requires=[
         "urwid>=1.1.1",
         "pygments>=1.0",
+        "jedi>=0.18,<1"
     ],
     test_requires=[
         "pytest>=2",


inducer:
Thanks for the PR! There was some discussion in #414 on whether Jedi is too heavyweight/slow, but maybe a sane approach here is to just drop it in and wait for people to complain. I've certainly been annoyed by not having Tab completion upon install in pudb, so to me this is a win.

hexagonrecursion:
> Jedi is too heavyweight/slow

1. I have not noticed
2. It is my understanding jedi is currently the only implementation of tab completion supported by pudb. Even if it is slow the only alternative is no completion at all

asmeurer:
I might have mentioned this on another issue, but it would be possible to extract the completion code from rlcompleter in a way that avoids importing readline (basically copy paste the code and delete the `import readline`). But I would only do that if people complain about jedi. Jedi already does `dir` completion, which is what rlcompleter does. 

By the way, when I added jedi support I didn't add it as a dependency because it's optional. It is only needed if you want to tab complete in the console, in which case it will give an error if jedi isn't installed. But I don't feel strongly about whether or not it should actually be a dependency, so if you feel the user experience is better this way we can just do this.

PR #435, opened 2021-03-24
Pylint, actually drop Py2 compatibility

inducer:
- [x] Gitlab sister MR: https://gitlab.tiker.net/inducer/pudb/-/merge_requests/14

@mvanderkamp This drops a bunch of py2 compat code, which I recall you were looking to preserve. How upset would you be if this got merged?


pudb/py3compat.py
```diff
@@ -2,27 +2,20 @@
 import sys
```
asmeurer, line 2: If we're getting rid of all this we can just delete this file, and the `__future__` imports everywhere. It's better to just use Python 3 names rather than replacing them with the old Python 2 names.

inducer: Fair enough. 7f1b08a

inducer:
@mvanderkamp @asmeurer Could one of you look over this to see if I've screwed up somewhere?

mvanderkamp:
> * [x]  Gitlab sister MR: https://gitlab.tiker.net/inducer/pudb/-/merge_requests/14
> 
> @mvanderkamp This drops a bunch of py2 compat code, which I recall you were looking to preserve. How upset would you be if this got merged?

Not upset at all- I've got my own branch sitting off in a corner that suits my python 2 purposes [here](https://github.com/mvanderkamp/pudb/tree/legacy-py2). It's a "no-promises" branch that I've cherry picked my changes onto as a testing ground for how they feel but otherwise haven't been keeping up to date. If other people are stuck in the same boat as me they're welcome to use it or fork it. FWIW I've been doing my main work on PRs in my home environment with python 3.

mvanderkamp:
I'll take another look once I'm in a proper env to test out the changes and see if anything is missing. So far the changes look sensible to me. Love the use of format strings and the general simplification that is going on here!

pudb/lowlevel.py
```diff
@@ -99,24 +96,15 @@ def _init_loggers():
...
-            lineno += line_incr
-            yield lineno
-    else:
-        for c in code.co_lnotab[1::2]:
-            lineno += ord(c)
-            yield lineno
+    # See https://github.com/python/cpython/blob/master/Objects/lnotab_notes.txt
+
+    for line_incr in code.co_lnotab[1::2]:
+        # This showed up between the v3.5 and v3.6 cycles:
```
inducer, line 102:
```suggestion
        # NB: This code is specific to Python 3.6 and higher
```

pudb/debugger.py
```diff
@@ -1743,23 +1728,15 @@ def cmdline_exec(w, size, key):
...
-                from io import StringIO
-            else:
-                from cStringIO import StringIO
+            from io import StringIO
 
             sys.stdin = None
             sys.stderr = sys.stdout = StringIO()
             try:
                 # Don't use cmdline_get_namespace() here in Python 2, as it
                 # breaks things (issue #166).
```
mvanderkamp, line 1737:
```suggestion
```
This comment is outdated.

mvanderkamp (changes requested):
I can't figure out how to add a comment to lines that weren't affected by the PR in github :(
Anyway there are several lookups of `sys.version_info` that appear to be checking for python 2:

- `example_stringifier.py` line 61
- `pudb/remote.py` line 107
- `pudb/__init__.py` line 73
- `setup.py` line 8
  - The PY_VERSION constant can probably be done away with here too.
  - Should there be a plain `pudb` console script entry point still though, in addition to `pudb3` or does that muddy the waters?
    - `doc/starting.rst` around line 30 discusses `pudb` and `pudb3` which should probably be cleaned up accordingly.

The comment for `_runmodule` in `debugger.py` refers to Python 2 still.

pudb/debugger.py
inducer: Thanks for taking a look! I think 71ffa51 should address all the points you raised. As for `pudb3`, I've left that in, but I've not added a `pudb` script back. I think that, ultimately, `python -m pudb` is the more robust thing to use. Ive' also changed the docs to push people towards that.

> I can't figure out how to add a comment to lines that weren't affected by the PR in github :(

I don't think Github allows it.

inducer:
I'll go ahead and merge this, since it seems like it is the way forward. I'll wait a bit with the next release though, in case any issues crop up. Thanks to the both of you for taking a look!

PR #438, opened 2021-04-02
Make the type stringifier just call type(), convert old type stringifier to a "default" stringifier

mvanderkamp:
This should address #384 as the logic for handling the "basic" types is now wrapped up in the "default" stringifier, which means you can now use other stringifiers on things like strings, ints, and booleans.


inducer:
Thanks for your contribution! I eliminated a redundant constant in 290f5cd, hope that's OK.

PR #441, opened 2021-04-17
Improve the README file - Add new screenshot and recent screencast link

ranelpadon:
This is the debugger I'm looking for for a long time since I switched from PyCharm/VS Code to Vim/Neovim. I use Pdb++ most of the time now, but this PuDB is a convincing alternative or much better replacement. I think I came across PuDB before in the past but was greatly repelled by the "awful" default screenshot in the GitHub page which I thought could not be customized. I only reconsidered this recently because someone commented about the [vim-pudb](https://www.reddit.com/r/vim/comments/jnkvfu/dear_vim_users_what_is_your_best_setup_for/gb3majr) plugin.

So, I searched YouTube for recent tutorials, fortunately the top result has a good theme in its video thumbnail!
https://www.youtube.com/watch?v=bJYkCWPs_UU

And that video turned out to be a concise and practical demo of PuDB. And here's a recent one too, and the **first question** after the talk was "if it's possible to change/customize the theme":
https://www.youtube.com/watch?v=_CzJ5bdRDU4&t=15m48s

The point of this PR is that color scheme is very important to programmers, one of the reasons also why we have UnixPorn/VimPorn subreddits. So, including a non-default theme in the **README** file would be a selling point for some devs which are repelled by the default theme, and could give them good first impressions instead. I updated the screencasts' links and emphasized the themes in the **Features** section as well. 

By the way, thanks for this powerful app! :)


README.rst
```diff
@@ -17,11 +17,20 @@ right where you write and test it--in a terminal. If you've worked with the
 excellent (but nowadays ancient) DOS-based Turbo Pascal or C tools, PuDB's UI
 might look familiar.
 
-Here's a screenshot:
+Here are the screenshots:
```
inducer, line 20:
```suggestion
Here are some screenshots:
```

inducer (commented):
Thanks for taking the time to improve the README! One minor suggestion (which I'll apply), otherwise this looks good to go.

inducer:
Hmm, turns out the new README looks terrible, the screenshot is way too large. I'll go ahead and fix that.

inducer:
#442

ranelpadon:
@inducer Actually, I matched the width of the first screenshot, but since they have different aspect ratio, the second screenshot looked too small, so I tried to match their height instead.

inducer:
No worries. My main concern is how off they looked on my Gitlab page: https://gitlab.tiker.net/inducer/pudb/-/tree/68808b8e053500da34672682860814023d6fde27

I still appreciate the nudge to refresh that ages-old screenshot.

ranelpadon:
@inducer Good idea also to put a representative Light and Dark themes in #442 that would be useful for newcomers. Cheers

PR #442, opened 2021-04-17
Fix README screenshot situation

inducer:


inducer:
Still not ideal on Gitlab:
https://gitlab.tiker.net/inducer/pudb

ranelpadon:
@inducer Looks good in GitHub though, which I think is more important:
https://github.com/inducer/pudb

PR #443, opened 2021-04-17
Give the correct length of remaining mapping items to expand in the variable view

mvanderkamp:
Bit of an oversight on my part, that comes into play with Pandas in particular but probably others as well. The "X items omitted" message was wrong for DataFrames: the length of a dataframe is the number of rows, but its keys are the columns. I changed it so that the length of the mapping's keys is checked instead of just the length of the mapping.

There's also a commit in here to make the "default" stringifier the actual default (d'oh).

### DIFF
--- a/pudb/settings.py
+++ b/pudb/settings.py
@@ -107,7 +107,7 @@ def load_config():
 
     conf_dict.setdefault("current_stack_frame", "top")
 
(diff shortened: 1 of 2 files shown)


inducer:
Thanks!

PR #447, opened 2021-04-29
Drop codecov

inducer:
https://about.codecov.io/security-update/

pudb wasn't affected AFAICT (not using the bash uploader), but I don't think the service is worth the potential supply chain attack vector. (I have some automatic integration with my private Gitlab instance set up, and I had to rotate credentials on ~30 repos in response to this. Fun.)


asmeurer:
+1. I've never found codecov useful (on any repo). Even when I care about coverage, it's not really great at showing it.

Issue #289, opened 2018-01-14
Remote debugger leaves its socket open


inducer:

sogaiu:
Current work-around here is to do:

```python
pudb.remote.debugger().do_quit(1)
```

after quitting a debugging session with pudb.

This made it possible to successfully start a new session and connect to it -- though depending on the situation may be on a higher port number.

inducer:
Thanks! I'd be happy to take a patch that makes this "just work".

Munk801:
This is great.  Glad to know I can at least close this as my terminal was getting blown if I just quit.

inducer:
#450 should have fixed this. Please comment if not.

sogaiu:
@inducer I tried with b36b10e8905971c700ddb6b41bf6102ea0a310c4 and it seems to be working :+1:

PR #450, opened 2021-05-13
Fixed issue with closing the session in remote debugging

mlubimow:
Fix for https://github.com/inducer/pudb/issues/289.

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -186,7 +186,8 @@ class Debugger(bdb.Bdb):
             **kwargs):
 
         if self.is_remote and not hasattr(self, 'close_remote_session'):
-            raise RuntimeError('Remote debuggers need to implement close_remote_session')
+            raise RuntimeError(
+                'Remote debuggers need to implement close_remote_session')
 
         # Pass remaining kwargs to python debugger framework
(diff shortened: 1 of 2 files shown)


inducer:
LGTM! Thanks for your contribution.

mlubimow:
@inducer Actually I realised recently that it's current behaviour also is not fully correct. I was focused on fixing the existing code more than the logic itself, and later I had second thoughts that on continue we probably don't want to close the session, because when in normal remote mode (reverse=False) it might be inconvenient to be required to telnet again. Do you know if there is any hook that would allow us to run some code once the debugged process ends?

On the good side, with reverse=True its easy as its enough to add -k flag to nc (stty -echo -icanon && nc -l -p 6899 -k) to make continue to work as expected.

inducer:
> Do you know if there is any hook that would allow us to run some code once the debugged process ends?

Possibly `atexit`?

inducer:
> Actually I realised recently that it's current behaviour also is not fully correct.

Should we revert this in the meantime?

mlubimow:
@inducer Yeah atexit seems to be a way better idea, I will create another PR fixing this.

PR #452, opened 2021-05-16
Better way to handle sockets cleanup in remote modes

mlubimow:
Context: #450

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -179,15 +179,8 @@ OTHER DEALINGS IN THE SOFTWARE.
 # {{{ debugger interface
 
 class Debugger(bdb.Bdb):
-    is_active = True
-    is_remote = False
-
     def __init__(self, stdin=None, stdout=None, term_size=None, steal_output=False,
             **kwargs):
-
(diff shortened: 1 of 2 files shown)


mlubimow:
Fixing #289 again, now it makes more sense.

There is something that is bothering me about remote modes (not in scope of this ticket), but I've noticed that while stdin is redirected for the period of the session, stderr is not. Do you know what was the reason to do it this way? Because I guess it can be misleading sometimes.

Another thing to think about it whether stdin / err / out should still be redirected when user press continue, I'm not fully sure. I left this behaviour as it was, but I wonder if there is any case when it wouldn't be preferred.

inducer:
This is much better. Thanks again for making this happen!

inducer:
> Another thing to think about it whether stdin / err / out should still be redirected when user press continue, I'm not fully sure. I left this behaviour as it was, but I wonder if there is any case when it wouldn't be preferred.

It comes down to which shell you'd like debugger errors to end up in. I figured that the shell in which you started the debugger would probably be a solid choice. If anything does go wrong, redirecting debugger errors into a (potential black hole) socket IMO has the potential to be unhelpful.

PR #453, opened 2021-05-26
[fix] find HOME value when abstent using os.path.expanduser

Psycojoker:
Hello,

This fix might appear in a weird context for you, the situation is:

- using pudb
- via [pytest-pudb](https://github.com/wronglink/pytest-pudb/)
- while calling pytest via tox
- the configuration will be ignored because: [pytest won't pass down the $HOME variable by default](https://tox.readthedocs.io/en/latest/example/basic.html#passing-down-environment-variables)
- thus you end up having to redo your configuration at every startup (which can be a dozen times per day)
- thus this is very annoying and and this is actually extremely not intuitive to debug (like I've spent a few hours in the source code)

So to handle this situation I'm using the `os.path.expanduser("~")` way to still get the user's home (maybe there's a better way?)

But to be honest I'm not sure where this need to be fixed?

You can do this in every tox file but it's a bit annoying to modify it only for debugging:

```ini
[testenv]
setenv =
  HOME={homedir}
```

Or maybe [it should be fixed in pytest-pudb?](https://github.com/wronglink/pytest-pudb/pull/23)

Or maybe directly in pudb because this will be a broader fix for more people?

I'm not sure

Nevertheless thanks a lot for this debugger, it really help me a lot!

Kind regards,

### DIFF
--- a/pudb/settings.py
+++ b/pudb/settings.py
@@ -32,6 +32,7 @@ from pudb.lowlevel import (lookup_module, get_breakpoint_invalid_reason,
 
 # minor LGPL violation: stolen from python-xdg
 
+# see https://github.com/inducer/pudb/pull/453 for context
 _home = os.environ.get("HOME", os.path.expanduser("~"))
 xdg_data_home = os.environ.get("XDG_DATA_HOME",
             os.path.join(_home, ".local", "share") if _home else None)
--- a/pudb/settings.py
+++ b/pudb/settings.py
@@ -32,7 +32,7 @@ from pudb.lowlevel import (lookup_module, get_breakpoint_invalid_reason,
 
 # minor LGPL violation: stolen from python-xdg
 
-_home = os.environ.get("HOME", None)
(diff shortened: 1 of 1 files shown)


pudb/settings.py
```diff
@@ -32,7 +32,7 @@
 
 # minor LGPL violation: stolen from python-xdg
 
-_home = os.environ.get("HOME", None)
+_home = os.environ.get("HOME", os.path.expanduser("~"))
```
inducer, line 35:
```suggestion
# see https://github.com/inducer/pudb/pull/453 for context
_home = os.environ.get("HOME", os.path.expanduser("~"))
```

inducer:
Thanks for your contribution, and for the thorough explanation! What you describe makes sense to me, and I'm happy to accommodate the use case. I added a comment from the code back to this PR so that the rationale for the slightly weird-looking code sticks around. (5ae6579)

Psycojoker:
Thanks for the quick merge!

PR #455, opened 2021-06-08
Use bpython.embed instead of deprecated bpython.cli.main

hhoeflin:
Thanks for pudb - it is great. I like to use bpython with it, and I have been getting deprecation notices (see [bpython.cli.main](https://github.com/bpython/bpython/blob/d07a405fe6b4ac8c51f7603db113dd88ed6713ae/bpython/cli.py#L1944)). In addition, the old interface had issues with ANSI color sequences, and the `bpython.embed` does not.

The current way of using this should be similar to ptpython, i,.e. using the `bpython.embed`.

Just a 2 line change - see below.


inducer:
Thanks!

PR #461, opened 2021-07-01
Source formatter refactor

mvanderkamp:
It bugged me that `UrwidFormatter` was being redefined on every call to `format_source`, so I've refactored it out of the function. Does this seem sensible? Also included is a fix to "dunder" method detection logic and the addition of highlighting support for "magic" methods.

By the way, I've been thinking about making it easier to create custom themes by not really having any default highlighting- build out the inheritance map to cover all tokens, and in the default pallette just define all tokens as `None`.

### DIFF
--- a/pudb/source_view.py
+++ b/pudb/source_view.py
@@ -253,6 +253,7 @@ else:
         t.Literal: "literal",
         t.Name.Exception: "exception",
         t.Name.Function: "name",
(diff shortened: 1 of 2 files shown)


inducer:
LGTM, thanks!

inducer:
Regarding the "empty" default theme: I don't feel strongly. If you think it's helpful, go for it!

Issue #254, opened 2017-04-29
command line default height


viniciusarcanjo:
Looking into the settings I didn't find a way to increase the height of the command line window by using setting a weight modifier, something similar to the 'sidebar_width' setting variable..

Is there a way to workaround this? Often times I have to keep adjusting the height of the command line window and having this default setting would be great. 

Thank you for this awesome debugger.

inducer:
I don't think there is, but I'd be happy to take a patch. :)

viniciusarcanjo:
@inducer that'd be awesome if you could. When this patch is available just let me know, I can help testing. 

Thank you.

Andrew-Wichmann:
Hi, is there any update on this?

inducer:
If you submit a patch, yes. :)

PR #462, opened 2021-07-12
Save internal shell height in the settings

### LINKED ISSUE #449: Internal shell height should be saved in the settings
I think the default height of internal shell is too small. 
Thx~

qhuy4119:
Internal shell height will be saved in the settings, similar to sidebar.
fixes #449 #254


inducer:
Thanks!

PR #463, opened 2021-07-13
Add more shortcuts in internal shell

### LINKED ISSUE #386: Could you please add more shortcuts in internal shell such as M-backspace?
When I'm using the internal shell to debug such as print value of something, I always miss the shortcuts in my system shell or ipython/bpython such as 

1. C-a/e to beginning/end of the prompt (both are already implemented)
2. C-d to delete current character
3. Alt-backspace to delete a word backward
4. C-u/k to delete to the beginning/end of the prompt(C-u is already implemented)

Could you please add these shortcuts?

qhuy4119:
Add more shortcuts when working with the internal shell using the
`urwid_readline` library. These shortcuts are the common default shortcuts
in system shell or ipython.

fixes #386


inducer:
Thanks!

PR #464, opened 2021-07-13
Display path of current source file in header bar

qhuy4119:
Currently, the full path of the current source file will be displayed.
This usually results in long path. When the path can't be displayed fully in 1 row for whatever reason (the window is resized, not enough space from the beginning,...), after a Ctrl-L (hotkey for redraw screen), it will be wrapped automatically, creating more rows as needed.

We have multiple options to resolve this:
- Leave as is
- Only display the filename
- Calculate the length of the string somehow (from $COLUMNS, for example)
- Let user decide in config file
- Other....

Let's discuss this

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -2642,10 +2642,8 @@ class DebuggerUI(FrameVarInfoKeeper):
         self.current_exc_tuple = exc_tuple
 
         from pudb import VERSION
-        caption = [(None,
-            "PuDB %s - ?:help  n:next  s:step into  b:breakpoint  "
-            "!:python command line"
-            % VERSION)]
+        caption = [(None, "PuDB %s - ?:help - %s" %
+                    (VERSION, self.source_code_provider.get_source_identifier()))]
 
         if self.debugger.post_mortem:
             if show_exc_dialog and exc_tuple is not None:


qhuy4119:
I updated it to calculate the path string automatically to fit on 1 line

pudb/debugger.py
```diff
@@ -2617,10 +2617,25 @@ def interaction(self, exc_tuple, show_exc_dialog=True):
...
+            elif available_width > len(full_filename):
+                return full_filename
+            else:
+                trim_index = len(full_filename) - available_width
+                filename = full_filename[trim_index:]
+                first_dirname_index = filename.find(os.sep)
+                filename = filename[first_dirname_index + 1:]
+                return filename
+
+        caption = [(None, separator.join([info_string, get_source_filename()]))]
```
inducer, line 2638: Thanks for working on that! This isn't quite the right way to do this IMO. Instead, this should be done in the repaint method for a widget, where Urwid supplies the amount of space available.

qhuy4119: I need more guidance because I'm not sure what you are referring to.
 > this should be done in the repaint method for a widget

Are you talking about [urwid.Text.render](http://urwid.org/reference/widget.html#urwid.Text.render), or some repaint code in PuDB, or something else?

inducer: Yes, I'm talking about subclassing `urwid.Widget` or `Text` and implementing our own `render` method.

mvanderkamp:
A thought on this: an optional footer might be a better place. The header currently displays very useful information that helps new users get started with using the debugger, and I don't think we should lose that.

PR #465, opened 2021-07-14
Fix crash when expanding variables in post-mortem

### LINKED ISSUE #233: Periodically crashes
Traceback:

```
  File "/usr/bin/pudb3", line 11, in <module>
    load_entry_point('pudb==2017.1.1', 'console_scripts', 'pudb3')()
  File "/usr/lib/python3.6/site-packages/pudb/run.py", line 32, in main
    steal_output=options.steal_output)
  File "/usr/lib/python3.6/site-packages/pudb/__init__.py", line 127, in runscript
    dbg.interaction(None, sys.exc_info(), show_exc_dialog=False)
  File "/usr/lib/python3.6/site-packages/pudb/debugger.py", line 339, in interaction
    show_exc_dialog=show_exc_dialog)
  File "/usr/lib/python3.6/site-packages/pudb/debugger.py", line 2079, in call_with_ui
    return f(*args, **kwargs)
  File "/usr/lib/python3.6/site-packages/pudb/debugger.py", line 2307, in interaction
    self.event_loop()
  File "/usr/lib/python3.6/site-packages/pudb/debugger.py", line 2273, in event_loop
    toplevel.keypress(self.size, k)
  File "/usr/lib/python3.6/site-packages/pudb/ui_tools.py", line 96, in keypress
    result = self._w.keypress(size, key)
  File "/usr/lib64/python3.6/site-packages/urwid/container.py", line 1128, in keypress
    return self.body.keypress( (maxcol, remaining), key )
  File "/usr/lib64/python3.6/site-packages/urwid/container.py", line 2269, in keypress
    key = w.keypress((mc,) + size[1:], key)
  File "/usr/lib/python3.6/site-packages/pudb/ui_tools.py", line 96, in keypress
    result = self._w.keypress(size, key)
  File "/usr/lib64/python3.6/site-packages/urwid/container.py", line 1587, in keypress
    key = self.focus.keypress(tsize, key)
  File "/usr/lib64/python3.6/site-packages/urwid/container.py", line 1587, in keypress
    key = self.focus.keypress(tsize, key)
  File "/usr/lib/python3.6/site-packages/pudb/ui_tools.py", line 101, in keypress
    return handler(self, size, key)
  File "/usr/lib/python3.6/site-packages/pudb/debugger.py", line 789, in edit_inspector_detail
    fvi = self.get_frame_var_info(read_only=False)
  File "/usr/lib/python3.6/site-packages/pudb/var_view.py", line 536, in get_frame_var_info
    ssid = self.debugger.get_stack_situation_id()
  File "/usr/lib/python3.6/site-packages/pudb/debugger.py", line 342, in get_stack_situation_id
    return str(id(self.stack[self.curindex][0].f_code))
IndexError: list index out of range
```

qhuy4119:
Currently, `Debugger.stack` will be empty when entering post-mortem, leading to `IndexError` when user tries to expand variables in the Variables window.
This commit fixes that by restoring `Debugger.bottom_frame` to `Debugger.stack`.

fixes #233

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -403,6 +403,7 @@ class Debugger(bdb.Bdb):
         self.stack, index = self.get_shortened_stack(frame, tb)
 
         if self.post_mortem:
+            self.stack.append((self.bottom_frame, self.bottom_frame.f_lineno))
             index = len(self.stack)-1
 
         self.set_frame_index(index)


pudb/debugger.py
```diff
@@ -403,6 +403,7 @@ def interaction(self, frame, exc_tuple=None, show_exc_dialog=True):
         self.stack, index = self.get_shortened_stack(frame, tb)
 
         if self.post_mortem:
+            self.stack.append((self.bottom_frame, self.bottom_frame.f_lineno))
```
inducer, line 406: How does `pdb` handle this situation? `self.stack` here is something that's managed by `bdb`, I'm not sure we're at liberty to mess with it.

qhuy4119: After finishing the execution of the script, pdb enters [this finally block](https://github.com/python/cpython/blob/e5862f79c16e28f1ec51d179698739a9b2d8c1d2/Lib/bdb.py#L600) (kinda like our post-mortem mode). If at this point, we type `p <variable name>` on the command line, pdb will print a `NameError`, and the user continues with the pdb REPL. So currently our `IndexError` is similar to this `NameError`.

I haven't dig into how bdb manages the stack in this case. 

The problem with the current code is that if I have an object at the module level, like `a = list(range(100))` for example, I can't expand it to see what the elements are in post-mortem. This PR allows user to inspect data structures like that.

inducer: Instead of modifying data structures that aren't ours to modify, could you instead create a `@property` that provides a modified view of the stack, and then change the relevant spots of the code to use that view of the stack instead? Also make sure to add comments to explain why this is being done.

inducer:
This actually isn't right. Post-mortem mode also kicks in if the debuggee runs into an exception. Your proposed change appears to break inspecting the stack in that situation.

PR #468, opened 2021-07-17
Fix function argument and "self" highlighting

mvanderkamp:
There was a bug in the argument parser that prevented it from working on magic methods.

I've also re-enabled highlighting `self` under the "pseudo" style. It's not clear to me why this was disabled to begin with, though my best guess is that it had something to do with the argument parser bug producing apparently wonky results.


inducer:
Thanks!

PR #469, opened 2021-07-27
Simplify custom theming

mvanderkamp:
There's nothing quite like a nice custom theme to really make someone feel at home in a UI. This PR will make it a lot easier to customize the look of pudb by defining a near-complete inheritance tree of styles, such that only five styles need to be defined, everything else is icing in on the cake.

To demonstrate this capability, I added a simple "mono" theme using only 6 styles.

There are a handful of tangential fixes present here as well:
- fixed background highlighting in the breakpoints view not going all the way to the end of the box.
- better error handling when loading custom themes (and same logic applied to other instances where we load a custom module)
- Make sure the themes all properly define their background colour at all times
- Don't change the theme if loading a custom theme fails.

Some of these changes may have some slight effects on the existing themes. I'm working on an fresh coat of paint for all the existing themes, I can add that to this PR or wait for a separate PR.


mvanderkamp:
I'm not entirely happy with the hierarchy, just a heads up that I'll be modifying it.

inducer:
Ok, I'll hold off on looking in more detail then. I just got back from a trip and was working through my backlog of stuff.

mvanderkamp:
Okay I think this is ready for another set of eyes. I went ahead with a theme remaster, but I've tackled each theme in a separate commit so feel free to reject all or any of them. My primary goals in the remasters were:

1. Simplify
2. Consistent style for "focused" element
3. Avoid weird contrast
4. For themes where I could find an example of the theme outside the context of `pudb`, try to align the theme with what it looks like elsewhere.
5. Reduce the number of colors used for 256 color themes. Especially the monokai-256 one, a lot of the styles were copy-pasted from agr-256 and didn't really match up with the syntax colors.
6. Make monokai look as much like monokai-256 as possible with the common Xterm theme.

I was primarily using GNOME terminal with the Xterm palette (except for solarized, obviously), but I also tried out a variety of other built-in palettes to check if at least one of the themes was at least usable, and I do think that's the case. If nothing else, mono is now always available as a fallback.

pudb/theme.py
```diff
@@ -23,1023 +23,856 @@
...
+        return style
+    except KeyError:
+        parent_name = INHERITANCE_MAP[style_name]
+        style = replace(get_style(palette_dict, parent_name), name=style_name)
+        palette_dict[style_name] = style
+        return style
+
+
+def link(child: str, parent: str):
+    INHERITANCE_MAP[child] = parent
```
inducer, line 257: I'm not a huge fan of the globally stateful design in use here. (I.e. I'm not liking that things are being modified in-place, and I like even less that these things are global variables.) Could this be refactored so that nothing needs to be modified in place? I think that might be possible by having each theme provide an additional inheritance map that's "overlaid" over the default one. But I'd be happy to consider other designs as well.

inducer (commented):
Thanks for working on this! The theming certainly was due an overhaul, and I like the direction you've taken here. I'd prefer if the logistics were less stateful, but I think that can hopefully be achieved without too much effort.

inducer:
I should add that, other than the statefulness aspect, this PR looks great to me.

pudb/theme.py
mvanderkamp: Yeah sure! This approach was a kludge that I threw in when I realized that themes could interfere with each other. I like the idea of having a local map overlaid over the global default one, that should be easy to do.

mvanderkamp: Okay, I added a local dict of inheritance overrides that is looked up first, and which is local to `get_palette`

PR #470, opened 2021-08-14
Breakpoint view fixes

mvanderkamp:
I noticed that toggling a breakpoint from a different file caused the source in the current file to be highlighted as though the breakpoint was actually there instead. Also, sometimes we would just hit traceback in this scenario. I cleaned up the relevant logic so that shouldn't be a problem anymore.

To test:
* Run the `try_the_debugger.sh` script, open some other module with 'm', and set a breakpoint on a line in the first 60 lines of the file, and another on a line somewhere after line 60 (so there's no corresponding line in `debug_me`).
* Go back to the `debug_me` module and, using 'b' from the breakpoint view, try toggling those breakpoints in the other module.

I also added in the highlighting fix for the breakpoint view,  since I think my theme work is going to take a little while yet.

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -1172,25 +1172,24 @@ class DebuggerUI(FrameVarInfoKeeper):
 
         # {{{ breakpoint listeners
 
+        def set_breakpoint_source(bp):
+            bp_source_identifier = \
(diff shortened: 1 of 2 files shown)


inducer:
LGTM, thanks!

PR #471, opened 2021-09-03
Properly update views after enabling/disabling breakpoint from the edit dialog

mvanderkamp:
Makes sure that enabled state is immediately reflected accurately in both source and breakpoint views after toggling whether a breakpoint is enabled in the "edit breakpoint" dialog.


inducer:
Thanks!

Issue #460, opened 2021-07-01
Feature request: Customizable key bindings using a config file


cibinmathew:
Enable debugger.py to read the key bindings and action mappings from a config file.

From a quick look, I think below could be modified
https://github.com/inducer/pudb/blob/ba31011c8f491bd6c36b59636f93ebb297ccc180/pudb/debugger.py#L2060-L2075

inducer:
It's not hard to get the key bindings from somewhere configurable. I would probably not be supportive of having UI to configure that (too much clutter), but having them in the config file seems halfway reasonable.

If you'd like to work on that, I'd be OK considering a PR.

PR #472, opened 2021-09-26
Added configurable key bindings

cibinmathew:
Tried to implement **Configurable key bindings** https://github.com/inducer/pudb/issues/460
- currently added Window chooser keys which could be extended later.
- will add readme entries post initial review if any

@inducer Pls let me know initial review comments

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -69,18 +69,27 @@ Features
 Configuration
 -------------
 
-Override default key bindings in the settings file (see `faq`_).
-Add the bindings under mentioned section in the config file.
+=============================
+Override default key bindings
+=============================
+
+
+-  Configure in the settings file (see `faq`_).
+
+- Add the bindings under mentioned section in the config file (see `urwid key syntax <http://urwid.org/manual/userinput.html#keyboard-input>`_).
+
+- Only few actions are supported currently.
+
 .. _faq: doc/misc.rst#faq
 
-Only few actions are supported currently.
-`urwid key syntax <http://urwid.org/manual/userinput.html#keyboard-input>`
 
-- window chooser bindings
+
 
 .. code-block:: ini
 
     [pudb]
+    
+    # window chooser bindings
     hotkeys_breakpoints = B
     hotkeys_code = C
     hotkeys_stack = S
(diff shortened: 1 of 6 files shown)


README.rst
```diff
@@ -66,6 +66,19 @@ Features
 * Should work with Python 3.6 and newer. (Versions 2019.2 and older continue
   to support Python 2.7.)
 
+Configuration
+-------------
+
+Override default key bindings
```
inducer, line 72: This should explain that what you're supplying here is *partial* documentation of the config file. You need to explain more, in complete sentences. Where does the config file live? How does one change it? When do changes take effect?

inducer (commented):
Thanks for working on this. I think that, generally, this light-touch approach is OK. The documentation needs more work to be intelligible though.

README.rst
cibinmathew: I have added a link to faq to avoid duplication

cibinmathew: @inducer Pls check the changes

cibinmathew:
@inducer pls check the review changes done.

README.rst
```diff
@@ -66,6 +66,35 @@ Features
 * Should work with Python 3.6 and newer. (Versions 2019.2 and older continue
   to support Python 2.7.)
 
+Configuration
```
inducer, line 69: How come this is in the README and not in the documentation (under `doc/`)?

cibinmathew: @inducer Not sure If I understood correctly. but I dont see README.rst under doc/ in the main branch.

cibinmathew:
@inducer Done flake8 changes as well. Pls check

README.rst
inducer: My point is that I think the documentation (see rendered at https://documen.tician.de/pudb/, generated from the files under `doc/` via sphinx) is likely a better spot for this than the README (at least IMO).

cibinmathew: okay. I will take a look at how it could be added and generated. 
We can merge this now if it is okay and track these doc changes through a subsequent PR in the meanwhile I figure out generation of docs. Either of these works. Let me know

PR #473, opened 2021-09-30
Add a "header warning" style for warnings in the header

mvanderkamp:
This allows defining a general "warning" style that can be specialized for any of the specific uses. The others already have specific style (search not found, command line error).


inducer:
Thx!

PR #481, opened 2021-11-01
Declare compatibility with Python 3.7+

### LINKED ISSUE #480: Version 2021.2 No Longer Supports Python 3.6, but the PyPI Page Still Claims Support
I just downloaded the newest version of PuDB from PyPI and was attempting to run it in a Python 3.6 environment. I found that it now tries to import from `dataclasses` which is a 3.7+ module. The PyPI page still claims support for Python 3.6 though. Should the page now be updated?

RobertBlackhart:
Fixes #480. Version 2021.2 now requires the use of `dataclasses` which is present only in Python 3.7 and up.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,7 @@ setup(
     long_description=long_description,
     author="Andreas Kloeckner",
     author_email="inform@tiker.net",
-    python_requires="~=3.6",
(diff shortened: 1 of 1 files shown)


inducer:
See #482.

PR #486, opened 2021-11-16
Remote debugging fixes

inducer:
Still to do: Make better facilities to run a function and catch exceptions from it. Currently something like this is necessary.
```
    import pudb
    if MPI.COMM_WORLD.rank == 0:
        from pudb.remote import debugger
        dbg = debugger()
        try:
            dbg.runcall(main_f)
        except Exception:
            pudb.pm()

    else:
        main_f()
        from time import sleep
        sleep(1000)
```

cc @matthiasdiener


matthiasdiener:
fwiw, I have to use smth like this:

```python
    from mpi4py import MPI

    import pudb
    if MPI.COMM_WORLD.rank == 0:
        from pudb.remote import debugger
        dbg = debugger()
        try:
            dbg.runcall(main_f)
        except Exception:
            pudb.pm()

    else:
        try:
            main_f()
        finally:
            from time import sleep
            sleep(1000)
```

matthiasdiener:
> Still to do: Make better facilities to run a function and catch exceptions from it. 

What do you think of #496?

PR #491, opened 2021-11-25
Fix #426, change history path of ptpython

### LINKED ISSUE #426: ptpython  now use ~/.local/share/ptpython/history     instead of      ~/.ptpython/history
should  this linein  pudb/shell.py be changed? 
` history_filename = os.path.expanduser("~/.local/share/ptpython/history") `

Freed-Wu:
I have tested it can work.

### DIFF
--- a/pudb/shell.py
+++ b/pudb/shell.py
@@ -246,10 +246,18 @@ def run_ipython_kernel(globals, locals):
 # }}}
 
 
+def get_ptpython_history_file():
+    from argparse import ArgumentParser
+    from ptpython.entry_points.run_ptpython import get_config_and_history_file
+    parser = ArgumentParser()
+    parser.add_argument("--history_file")
+    parser.add_argument("--config_file")
+    return get_config_and_history_file(parser.parse_args())[1]
+
+
 def run_ptpython_shell(globals, locals):
     # Use the default ptpython history
-    import os
-    history_filename = os.path.expanduser("~/.ptpython/history")
+    history_filename = get_ptpython_history_file()
     ptpython_embed(globals=globals.copy(), locals=locals.copy(),
                    history_filename=history_filename,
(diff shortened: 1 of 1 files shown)


pudb/shell.py
```diff
@@ -249,7 +249,7 @@ class DummyMod:
 def run_ptpython_shell(globals, locals):
     # Use the default ptpython history
     import os
-    history_filename = os.path.expanduser("~/.ptpython/history")
```
mvanderkamp, line 252: This is still supported as the "legacy history file":
https://github.com/prompt-toolkit/ptpython/blob/d8b5c90d20cc1e8d88837aa7163289950c231f74/ptpython/entry_points/run_ptpython.py#L131

Ideally, we would look up the history file using ptpython's methods, or at least emulate them or allow users to configure it to the desired location.

Freed-Wu: How do you think of `os.access(os.path.expanduser("~/.ptpython/history"), os.W_OK)`?

Freed-Wu: If `~/.ptpython/history` don't exist, let us use `~/.local/share/ptpython/history`.

inducer: Why not just call `ptpython.get_config_and_history_file`, as @mvanderkamp suggests?

https://github.com/prompt-toolkit/ptpython/blob/d8b5c90d20cc1e8d88837aa7163289950c231f74/ptpython/entry_points/run_ptpython.py#L131

inducer: Thanks! Failed the linter, otherwise LGTM!

Freed-Wu:
I have remove unused `import os`. Thanks!

inducer:
LGTM, thanks!

inducer:
Replaced by #495.

PR #496, opened 2021-12-16
add debug_remote_on_single_rank

matthiasdiener:
Needs #486.

### DIFF
--- a/doc/starting.rst
+++ b/doc/starting.rst
@@ -1,5 +1,5 @@
-Getting Started
----------------
+Starting the debugger
+---------------------
 
 To start debugging, simply insert::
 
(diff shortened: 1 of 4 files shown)


inducer:
Thanks for making this. I think this is a bit too specific to mpi4py to be of broad use, especially with as little documentation as it currently has.

matthiasdiener:
> Thanks for making this. I think this is a bit too specific to mpi4py to be of broad use, 

Do you have a suggestion on how to improve it?

> especially with as little documentation as it currently has.

Do you mean the documentation specifically for `debug_remote_on_single_rank` or for the remote debugging feature in general?

inducer:
Sorry, I did not mean to close this. I think it got auto-closed because the base branch went away.

> Do you mean the documentation specifically for `debug_remote_on_single_rank` or for the remote debugging feature in general?

I guess I could be talked into adding something like this, as long as it's

- included in the docs, and
- it has sensible semantics and is well-explained in terms of what it does. (what happens to the other ranks? they spin. what does the `rank` argument do? allow passing a communicator)

PR #499, opened 2021-12-31
Prefer packaging for determining versions

alexfikl:
All of `distutils` is deprecated now and this `packaging` thing seems to be what people are porting to (from a quick search).


inducer:
Thanks!

PR #503, opened 2022-01-15
Miscellaneous touchups in the var view

mvanderkamp:
This is several minor commits, feel free to take or leave whichever you're interested in!
- Various small bits of cleanup in `var_view.py`
- Give a specific error message if a user accidentally presses `c` in the var view when they don't have a custom stringifier set
- Show then length of all sized items, not just the builtin containers.


inducer:
LGTM, thanks!

PR #505, opened 2022-02-05
Add pytest-mock to test_requires in setup.py

bcbnz:
This is already in requirements-dev.txt but should be here as well as some of the unit tests fail to run without its `mocker` fixture.


inducer:
Thx!

PR #506, opened 2022-02-13
Key-bindings: toogle_cmdline_focus [WIP]

cibinmathew:
Adding more to Configurable key bindings https://github.com/inducer/pudb/issues/460


inducer:
- Please look into the linter failures.
- Maybe #507 might be good to address as part of this.

PR #515, opened 2022-05-19
Fix stack oddities

inducer:
This fixes a few oddities about the stack window that have been irritating me for a while:

- In post-mortem mode, there were often two copies of the stack in the stack window, one from the backtrace and the other from the frame objects.
- Post-mortem stacks may contain the same frame twice (e.g. when an exception is caught and re-raised: once for the location of the exception in this frame, and once more for the handler). Thus: Identify stack frames by index, not frame object.

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -397,11 +397,13 @@ class Debugger(bdb.Bdb):
         else:
             tb = exc_tuple[2]
 
-        if frame is None and tb is not None:
-            frame = tb.tb_frame
+        if frame is None:
+            assert tb is not None
+            walk_frame = tb.tb_frame
+        else:
+            walk_frame = frame
 
         found_bottom_frame = False
-        walk_frame = frame
         while True:
             if walk_frame is self.bottom_frame:
                 found_bottom_frame = True
@@ -411,6 +413,7 @@ class Debugger(bdb.Bdb):
             walk_frame = walk_frame.f_back
 
         if not found_bottom_frame and not self.post_mortem:
+            # We aren't supposed to be debugging this.
             return
 
         self.stack, index = self.get_shortened_stack(frame, tb)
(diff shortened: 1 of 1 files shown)


pudb/debugger.py
```diff
@@ -2799,11 +2804,12 @@ def make_frame_ui(frame_lineno):
...
                     class_name = "!! %s !!" % message
 
-            return StackFrame(frame is self.debugger.curframe,
+            return StackFrame(i == self.debugger.curindex,
                     code.co_name, class_name,
                     self._format_fname(code.co_filename), lineno)
 
-        frame_uis = [make_frame_ui(fl) for fl in self.debugger.stack]
+        frame_uis = [make_frame_ui(i, fl)
+                for i, fl in enumerate(self.debugger.stack)]
```
mvanderkamp, line 2812: Nice! I recall encountering duplicated selection in the stack view in other contexts as well, this should fix all of that up!

doc/conf.py
```diff
@@ -23,5 +23,5 @@
 
 intersphinx_mapping = {
     "https://docs.python.org/3": None,
-    "urwid": ("https://urwid.org/", None),
+    "urwid": ("https://urwid.readthedocs.io/en/latest/", None),
```
mvanderkamp, line 26: When I follow the readthedocs link it shows me version 1.3.1, but the urwid.org link shows version 2.1.2. The github page for urwid links to urwid.org. Have they simply not updated their page on readthedocs in a while?

mvanderkamp (commented):
I haven't tested the behaviour, but the changes look good!

doc/conf.py
inducer: You're right. It looks like https://urwid.org/ is still the official source. That had TLS issues yesterday, which made me go look for an alternative. Removed.

inducer:
You know, let's just try this. If nothing else, I think this is a step in a reasonable direction.

Issue #511, opened 2022-04-02
Add the ability to set a watch expression on an object by ID


quodlibetor:
**Is your feature request related to a problem? Please describe.**

I would love to be able to trace the evolution of a value across stack frames. This means that variables don't exist, but often the *object* that I care about is being mutated.

**Describe the solution you'd like**

The nicest thing, I think, would be for the watch expression dialog to get an enhanced view (maybe a select box that changes it to be this view)

| Object Reference | Watch expression |
|-------------------|-------------------|
| `myobj________` | `myobj.value___`  |

In an ideal world, pudb would look up the `id(myobj)` and then every time it entered a frame it would scan the variables to see if any of them have the same ID as myobj. If they do, then:

* if the object exists and has the same name you get the current display `myobj.value: ...`
* if the variable has a different name you get: `(myobj -> newname).value: ...`
* if the object does not exist: `myobj: <missing>`

**Describe alternatives you've considered**

Currently I just constantly re-enter watch expressions in every frame, whether or not the variable name has changed. 

Having the option to have watch expressions persist across frames would be helpful, usually there is a set of at most 4 var names that I care about and I wouldn't mind having 3 of the 4 be an error at any given time if it meant I didn't need to re-enter expressions all the time.

inducer:
I can see the use case. I'd be happy to consider a PR. This could be done by simply storing a reference to an object instead of the name in a watch expression, maybe with a "by-name" vs "by-value" toggle in watch expression creation.

mvanderkamp:
I'm taking a look at this, and I see two potentially separate ideas here:

1. Whether the watch expression exists globally in all frames or only the current/local one. At the moment all watch expressions exist on in their "local" frame.
2. Whether the watch expression should be re-evaluated every time using the currently existing local/global namespaces, or should store a reference to the first evaluated result and always display that result- most likely because it is a reference to a variable that exists in the current scope.

So, some questions:

1. Is it reasonable to join those two ideas together, or should they be distinct? I.e. should a watch always be either a local frame expression or an "all frames" reference to a variable?
2. Is it a valid use case to want to create a watch expression that will persist and be re-evaluated in multiple frames against different variables that share the same name? (Perhaps if you're dealing with several instances of the same class across multiple methods?)
3. Should the references be `weakref` references? 
    a) Would this require a slightly adjusted UI flow where users have to write/select a currently available variable instead of an arbirtrary expression?
4. Should we try to preserve information about where the reference came from?
[...]

PR #525, opened 2022-06-29
Add global and "by reference" watches

mvanderkamp:
Addresses #511 

Allows modifying watch expressions in two ways:
1. Whether to show the watch expression in all frames as opposed to only the current frame
2. Whether to continually re-evaluate the watch expression or evaluate it when it is created, cache the result, and watch changes to this value over time. If the value appears in the current local or global namespaces, it's current name is shown in parentheses.


mvanderkamp:
Actually what do you think of making all watches "global"? That's usually how I want watches to behave.

mvanderkamp:
Hmm just as well I'm not entirely happy with the state of this- we don't have access to the right InspectInfo for global watches since it's still being cached at the frame level, so the global watches are presented differently (display type, show detail, etc) in each frame.

mvanderkamp:
Also, any thoughts on the UI for this stuff? I went with "expression"/"reference" since that made sense to me at the time but "by name"/"by value" might make more sense. And how cumbersome is it to have potentially 4 different kinds of watches?

raphCode:
Haven't read up all the watch-related stuff, but here is a late night thought:
What about showing (some) watch values only when their evaluation does not fail? Put differently, hiding all watches that yield a `WatchEvalError`.

This is what comes to mind after seeing the behavior of https://github.com/inducer/pudb/pull/150, which saves watch expressions and loads them in every entered frame, which of course fails in many cases.

PR #530, opened 2022-07-19
refactor: replace LGPL code

rul:
This contribution replaces code marked as LGPL to avoid conflicts with the MIT license used in this project. Moreover, it simplifies the code by removing the `*resource` parameter in the `get_save_config_path` function, given that it isn't used in this project. It also removes the `xdg_data_home` global variable, as it isn't in use either.

For context, in Debian [we need to resolve this violation](https://bugs.debian.org/1005854) to keep the package in the archive.

### DIFF
--- a/pudb/settings.py
+++ b/pudb/settings.py
@@ -30,16 +30,12 @@ from configparser import ConfigParser
 from pudb.lowlevel import (lookup_module, get_breakpoint_invalid_reason,
                            settings_log)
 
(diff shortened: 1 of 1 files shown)


inducer:
Thanks! I'll roll a release to make this easier for you to package.

inducer:
https://pypi.org/project/pudb/2022.1.2/

PR #533, opened 2022-07-27
Fix critical bug when re-entering IPython 8.0.0 (and later) external shell

vsydorov:
I've found a bug with recent IPython 8.0.0 shell, which disrupts my personal workflow. Opening the external IPython shell (via !) only works properly the first time, namespace gets disrupted afterwards. You can verify this by calling globals() within the IPython shell and comparing results after the first and second time you enter the shell.

Here's a little fix for this bug:
- IPython 8.0.0+ performs at-exit operations after calling mainloop
  (introduced by 08d54c0e367b535fd88aca5273fd09e5e70d08f8 commmit),
  breaking necessary PuDB functionality. This prevents re-entering
  IPython shell
- Quick fix sets/unsets internal shell._atexit_once_called property
  during run_ipython_shell_v11, preventing those operations from being
  executed

### DIFF
--- a/pudb/shell.py
+++ b/pudb/shell.py
@@ -220,7 +220,14 @@ def run_ipython_shell_v11(globals, locals):
         args.append(banner)
     else:
         print(banner)
+
+    # XXX Quick and dirty way to prevent issues with IPython 8.4.0+, introduced by commit 08d54c0e367b535fd88aca5273fd09e5e70d08f8
+    # Setting this property will prevent call to IPython.core.interactiveshell.InteractiveShell._atexit_once()
+    # from inside IPython.terminal.interactiveshell.TerminalInteractiveShell.mainloop()
+    # This allows us to repeatedly re-call mainloop() and the whole run_ipython_shell_v11() function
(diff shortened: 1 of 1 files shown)


inducer:
Thanks! This isn't beautiful, but I'd be OK with merging it.
- Could you look at the linter failure?
- Do you have a sense whether this impacts older versions of IPython in any way?

vsydorov:
Hi, I've fixed E:501 that was causing that linter failure. Also made edits to reflect that the bug was introduced in version 8.0.0 of IPython. I went down the version history of IPython, up to the last supported  (7.10.0) by current pudb version. Didn't notice any issues caused by the proposed fix.

Versions checked:
- 8.5.0.dev (latest 66aeb3fc55c8ac04242e566172af5de5cc6fe71e commit from July 18)
- 8.4.0, 8.3.0, 8.2.0, 8.1.0, 8.0.0 
- 7.32.0, 7.15.0, 7.10.0

inducer:
Thanks for taking a thorough look!

PR #534, opened 2022-07-28
Close some files

inducer:


marc-h38:
This is causing regression #598

PR #535, opened 2022-07-29
Fix test_executable_lines for Python 3.11

frenzymadness:
See changes in CPython
https://github.com/python/cpython/pull/94562
https://peps.python.org/pep-0626/


inducer:
Thanks! This looks good, but could you address the linter failure?

frenzymadness:
Should be okay now.

inducer:
Thanks!

PR #536, opened 2022-07-29
Get terminal size using os.get_terminal_size()

h5rdly:
Hi,

A small PR to swap out the usage of `fcntl` / `termios` with `os.get_terminal_size()`. 

I was hoping to try and help make `pudb` cross-platform, this looks like a first step :)

Thanks,
Eli

### DIFF
--- a/pudb/forked.py
+++ b/pudb/forked.py
@@ -1,6 +1,5 @@
 import sys
 import os
-import struct
 
 from pudb.debugger import Debugger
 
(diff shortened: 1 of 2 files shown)


inducer:
Thanks! This looks good to me, other than the linter failure. Could you take a look?

AFAICT, `get_terminal_size` is available on Py3 unconditionally, given the lack of version numbers [here](https://docs.python.org/3/library/os.html#os.get_terminal_size).

inducer:
Thanks!

h5rdly:
My bad, removed unused `import struct`.

`get_terminal_size()` works on `3.6`+, can't vouch for earlier versions :)

PR #554, opened 2022-09-13
Fix zsh completion of `--pre-run`

Freed-Wu:
Refer https://github.com/zsh-users/zsh/blob/master/Completion/Unix/Command/_sudo#L47-L51
the original completion of `--pre-run` will complete external commands, environment variables, built-in functions of zsh, etc.
Now it only complete external commands, which is expected result of `--pre-run`: it cannot run built-in functions of shell.

### DIFF
--- a/pudb/run.py
+++ b/pudb/run.py
@@ -1,4 +1,4 @@
-COMMAND = {"zsh": "_command_names -e"}
+COMMAND = {"zsh": "{_command_names -e}"}
 PREAMBLE = {
     "zsh": """\
 _script_args() {
@@ -23,10 +23,11 @@ def get_argparse_parser():
     version_info = "pudb:%(prog)s v" + VERSION
 
     if sys.argv[1:] == ["-v"]:
-        print(version_info % {"prog": os.path.basename(__file__)})
+        print(version_info % {"prog": "pudb3"})
         sys.exit(os.EX_OK)
 
     parser = argparse.ArgumentParser(
+        "pudb3",
         usage="%(prog)s [options] [-m] SCRIPT-OR-MODULE-TO-RUN [SCRIPT_ARGS]",
         epilog=version_info
     )
--- a/pudb/run.py
+++ b/pudb/run.py
@@ -20,14 +20,14 @@ def get_argparse_parser():
 
     from pudb import VERSION
 
-    version_info = "pudb:%(prog)s v" + VERSION
+    version_info = "%(prog)s v" + VERSION
 
     if sys.argv[1:] == ["-v"]:
-        print(version_info % {"prog": "pudb3"})
+        print(version_info % {"prog": "pudb"})
         sys.exit(os.EX_OK)
 
     parser = argparse.ArgumentParser(
-        "pudb3",
+        "pudb",
         usage="%(prog)s [options] [-m] SCRIPT-OR-MODULE-TO-RUN [SCRIPT_ARGS]",
         epilog=version_info
     )
--- a/setup.py
+++ b/setup.py
@@ -53,7 +53,7 @@ setup(
     entry_points={
         "console_scripts": [
             # Deprecated. Should really use python -m pudb.
-            "pudb3 = pudb.run:main",
+            "pudb = pudb.run:main",
             ],
         "gui_script": [],
     },
--- a/test/test_run.py
+++ b/test/test_run.py
@@ -30,7 +30,7 @@ def test_main_version(capsys, mocker, argv):
 
     captured = capsys.readouterr()
 
-    assert "pudb:pudb3 v" in captured.out
+    assert "pudb v" in captured.out
 
 
 def test_main_v_with_args(capsys, mocker):
--- a/test/test_run.py
+++ b/test/test_run.py
@@ -30,7 +30,7 @@ def test_main_version(capsys, mocker, argv):
 
     captured = capsys.readouterr()
 
-    assert "pudb:run.py v" in captured.out
+    assert "pudb:pudb3 v" in captured.out
 
 
 def test_main_v_with_args(capsys, mocker):


pudb/run.py
```diff
@@ -1,9 +1,15 @@
-COMMAND = {"zsh": "_command_names -e"}
+COMMAND = {"zsh": "_pre_run"}
 PREAMBLE = {
     "zsh": """\
 _script_args() {
   _arguments -S -s '(-)1:script_args:_files -g "*.py"' '*: :_files'
 }
+
+_pre_run() {
+  # https://github.com/zsh-users/zsh/blob/master/Completion/Unix/Command/_sudo#L47-L51
```
inducer, line 9: `master` could change. Please make this a versioned reference.

pudb/run.py
```diff
@@ -1,9 +1,15 @@
...
+  _comp_priv_prefix=($words[1] -n ${(kv)opt_args[(I)(-[ugHEP]|--(user|group|set-home|preserve-env|preserve-groups))]})
```
inducer, line 10: I'm a little lost here. What does this mean/do?

pudb/run.py
Freed-Wu: Fixed. :smile:

pudb/run.py
Freed-Wu: See the commit message of <https://github.com/zsh-users/zsh/commit/359e0dae2bded75eebf7fe3b3cf919e1dd44c333#diff-b3f104490c12fc9f1ef7365c016b1b51c1cc409454417bec00f98e96f6d1f44a>:

```
…sudo

* Update _command_names to add **/sbin to PATH when not otherwise overridden
  and _comp_priv_prefix is set. This ensures that, on systems where
  unprivileged users don't have **/sbin in their PATHs, we can still complete
  those commands when they try to run them with e.g. sudo

* Update functions for privilege-gaining commands to use the aforementioned
  mechanism

* Fix some other minor issues in _pfexec and _doas

Changes from the posted diff:

* Fix a minor redundancy in _doas

* Add _c_p_p=( '' ) hack to _su
```

pudb/run.py
```diff
@@ -27,6 +33,7 @@ def get_argparse_parser():
         sys.exit(os.EX_OK)
 
     parser = argparse.ArgumentParser(
+        'pudb3',
```
Freed-Wu, line 36: I notice `python -m pudb --print-completion zsh` will generate `#compdef __main__`. This line can make it always `#compdef pudb3`

Freed-Wu:
Now all checks have passed.

pudb/run.py
```diff
@@ -22,11 +26,8 @@ def get_argparse_parser():
 
     version_info = "pudb:%(prog)s v" + VERSION
 
-    if sys.argv[1:] == ["-v"]:
-        print(version_info % {"prog": os.path.basename(__file__)})
-        sys.exit(os.EX_OK)
```
inducer, line 27: How come this was removed?

pudb/run.py
```diff
@@ -1,9 +1,15 @@
...
+COMMAND = {"zsh": "_pre_run"}
 PREAMBLE = {
     "zsh": """\
 _script_args() {
   _arguments -S -s '(-)1:script_args:_files -g "*.py"' '*: :_files'
 }
+
+_pre_run() {
+  # https://github.com/zsh-users/zsh/blob/a9386df5581487e5e34ecb3ce422bacf62d10c3d/Completion/Unix/Command/_sudo#L47-L51
+  local _comp_priv_prefix=($words[1] -n ${(kv)opt_args[(I)(-[ugHEP]|--(user|group|set-home|preserve-env|preserve-groups))]})  # noqa: E501
```
inducer, line 10: These flags look like they're specific to `sudo`?

Freed-Wu:
```man
_command_names [ -e | - ]
      This  function completes words that are valid at command position: names of aliases, builtins, hashed commands, functions, and so on.  With the -e flag, only hashed commands are
      completed.  The - flag is ignored.

_comp_priv_prefix
      Completion functions such as _sudo can set the _comp_priv_prefix array to a command prefix that may then be used by _call_program to match the privileges when  calling  programs
      to generate matches.
```

Right. So I remove `_comp_priv_prefix`. I found the true reason is
`_command_names -e` must be wrapped in a function, although I don't know why his 
phenomenon happen.

## Before

Don't wrap `_command_names -e`, it will complete 8790 items and containing 
parameters like `USER`, `USERNAME`, `OSTYPE` which we don't need.

```zsh
  "--pre-run[Run command before each program run]:pre_run:_command_names -e"
```

```sh
❯ pudb3 --pre-run <TAB>
zsh: do you wish to see all 8970 possibilities (6300 lines)? y
❯ pudb3 --pre-run
pre_run
external command
parameter
7za                                 A file archiver with highest compression ratio
7z                                  A file archiver with highest compression ratio
7zr                                 A file archiver with highest compression ratio
a2ping                              convert between PS, EPS and PDF and other page description formats
a52dec                              decode ATSC A/52 audio streams
aafire                              aalib example programs
ab                                  Apache HTTP server benchmarking tool
aclocal-1.16                        manual page for aclocal 1.16.5
...
```

## After

Wrap `_command_names -e` to a function then call this function, it will complete 
2927 items and only contain external commands.

```zsh
  "--pre-run[Run command before each program run]:pre_run:_pre_run"
...
_pre_run() {
  _command_names -e
}
```

```sh
❯ pudb3 --pre-run <TAB>
zsh: do you wish to see all 5011 possibilities (2927 lines)? y
❯ pudb3 --pre-run
external command
7za                                 A file archiver with highest compression ratio
7z                                  A file archiver with highest compression ratio
7zr                                 A file archiver with highest compression ratio
...
```

PR #559, opened 2022-09-25
Implement "--continue" CLI arg (alias "-c")

mvanderkamp:
Adds a `--continue` (alias `-c`) CLI option. To my knowledge we don't have a command language like `pdb` does, so I thought it would be simpler to go about supporting this feature this way, as oppose to a generic `-c` like `pdb` offers. Alternatively I suppose we could call the keyboard controls a "command language" and try to find a way feed a series of keyboard inputs through.

### DIFF
--- a/pudb/__init__.py
+++ b/pudb/__init__.py
@@ -106,8 +106,8 @@ def runmodule(*args, **kwargs):
 
 
 def runscript(mainpyfile, args=None, pre_run="", steal_output=False,
-              run_as_module=False):
-    dbg = _get_debugger(steal_output=steal_output)
+              _continue=False, run_as_module=False):
+    dbg = _get_debugger(steal_output=steal_output, _continue=_continue)
 
     # Note on saving/restoring sys.argv: it's a good idea when sys.argv was
     # modified by the script being debugged. It's a bad idea when it was
--- a/pudb/__init__.py
+++ b/pudb/__init__.py
@@ -106,8 +106,8 @@ def runmodule(*args, **kwargs):
 
 
 def runscript(mainpyfile, args=None, pre_run="", steal_output=False,
-              _continue=False, run_as_module=False):
-    dbg = _get_debugger(steal_output=steal_output, _continue=_continue)
+              _continue_at_start=False, run_as_module=False):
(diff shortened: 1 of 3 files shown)


pudb/__init__.py
```diff
@@ -106,8 +106,8 @@ def runmodule(*args, **kwargs):
 
 
 def runscript(mainpyfile, args=None, pre_run="", steal_output=False,
-              run_as_module=False):
-    dbg = _get_debugger(steal_output=steal_output)
+              _continue=False, run_as_module=False):
```
inducer, line 109: I think it'd be nicer if these were also called `_continue_at_start` or so.

pudb/debugger.py
```diff
@@ -472,6 +478,10 @@ def user_return(self, frame, return_value):
                 return
             self._wait_for_mainpyfile = False
             self.bottom_frame = frame
+            if self._continue_at_start:
+                self._continue_at_start = False
+                self.set_continue()
+                return
```
inducer, line 484: Would it make sense to factor out the `_wait_for_mainpyfile` logic?

inducer (commented):
Thanks! Two comments below, otherwise LGTM.

asmeurer:
I also prefer this solution. I haven't seen anyone requesting a "command language" beyond this specific feature request. I'd like to see an actual use case for it before attempting something like that.

pudb/__init__.py
mvanderkamp: Fair point! I had to come up with a different name for the permanent value, it came out a bit awkward but hopefully it's clear at least. I wonder if it would make sense to capture that sort of thing in a settings object instead though.

pudb/debugger.py
mvanderkamp: done!

mvanderkamp:
Yeah I much prefer things this way too, I just figured I should point out the difference.

pudb/__init__.py
inducer: Awkward for sure, but I've seen worse. This works for me.

inducer:
Thanks for working on this!

PR #568, opened 2022-11-07
Improve sidebar help

kwmiebach:
Being more specific about which keys adjust width or height. See also https://stackoverflow.com/a/64482286


inducer:
Thanks!

PR #574, opened 2022-11-18
Show shell error messages in fallback shell

### LINKED ISSUE #573: "Falling back to custom shell" message printed in internal shell console
When the custom shell is not installed, pudb automatically fallsback to the classic Python shell. But the message for this is printed in the internal shell console. This is useless because this isn't shown when the shell is active. It should be printed to the terminal. I think this used to be the case, so this likely was changed at some point.

mvanderkamp:
Also continue to show the error message in the ui_log.

Fixes #573


inducer:
Thanks for working on this!

PR #586, opened 2023-01-20
Fix #580, Source $PYTHONSTARTUP before entering REPL

Freed-Wu:
If users directly enter REPL from command line, $PYTHONSTARTUP will be
sourced. However, XXpython_embed() will not source $PYTHONSTARTUP.
Similarly, maybe we can provide a command line option -E to disable
this behavior, just like python itself.

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -24,6 +24,7 @@ THE SOFTWARE.
 """
 
 
+import logging
 import urwid
 import bdb
 import gc
@@ -39,6 +40,8 @@ from pudb.settings import load_config, save_config, get_save_config_path
 
 CONFIG = load_config()
 save_config(CONFIG)
+
+logger = logging.getLogger(__name__)
 
 HELP_HEADER = r"""
 Key Assignments: Use Arrow Down/Up or Page Down/Up to scroll.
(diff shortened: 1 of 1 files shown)


pudb/debugger.py
```diff
@@ -2109,7 +2113,16 @@ def fallback(error_message):
                         else:
                             runner = shell.custom_shell_dict["pudb_shell"]
 
-                runner(curframe.f_globals, curframe.f_locals)
+                pythonstartup = os.getenv("PYTHONSTARTUP", os.devnull)
+                f_globals = deepcopy(curframe.f_globals)
+                f_locals = deepcopy(curframe.f_locals)
```
inducer, line 2118: That won't fly. We want to be able to retain the ability to modify locals. At the same time, it's not great to have `$PYTHONSTARTUP` modify the local frame. I'm not sure there's a good way forward here.

Freed-Wu: So should we remove `deepcopy()`?

Freed-Wu: Removed.

<https://github.com/inducer/pudb/discussions/580#discussioncomment-4801771> seems to be a better solution.

Freed-Wu:
Looks like <https://github.com/inducer/pudb/discussions/580#discussioncomment-4801771> is better. So I close this issue now. Any thought can be commented to let me know :smile:

Issue #607, opened 2023-06-19
"Debugger instance already exists" when used in pytest


hydrargyrum:
# Describe the bug
When 2 breakpoints are encountered in pytest with pudb set up, this error pops: Debugger instance already exists

# To Reproduce
1. install pudb and pytest in venv
2. create this `test.py`
```
def test_foo():
    breakpoint()
    breakpoint()
```
4. `pytest --pdbcls pudb.debugger:Debugger --pdb --capture=no test.py` (following https://documen.tician.de/pudb/starting.html#usage-with-pytest)
5. First breakpoint is encountered in pudb, press `c` to continue to second one
6. See error

# Expected behavior
The debugger should be called on second breakpoint

# Versions

pudb 2022.1.3

inducer:
Why not use `pytest-pudb`? (installable via `pip install`)

hydrargyrum:
Actually, it's not working very well either. After installing it and running `pytest --pudb test.py --pdbcls pudb.debugger:Debugger`, I get the same error, but within pudb this time

inducer:
Try without the `--pdbcls`.

souliane:
I can't reproduce with 2 `breakpoint()`, but with this instead:
```
def test_foo():
    breakpoint()
    raise Exception
```
And running with the same command: `pytest --pdbcls pudb.debugger:Debugger --pdb --capture=no test.py`
It fails with:
```
  File "[...]/.venv/lib/python3.11/site-packages/pudb/debugger.py", line 196, in __init__
    raise ValueError("a Debugger instance already exists")
ValueError: a Debugger instance already exists
```
I expect to see this instead (here I pressed already `e`):
![image](https://github.com/inducer/pudb/assets/21343492/a74a8fdd-d835-42e0-ab39-a093aedf9b5d)

The issue can be fixed by commenting these two lines [here](https://github.com/inducer/pudb/blob/main/pudb/debugger.py#L190):
```
        if Debugger._current_debugger:
           raise ValueError("a Debugger instance already exists")
```
But I assume that this fix is too dirty to be committed... yet I don't understand why this check is needed, for me it works fine (better) without it.

Issue #52, opened 2013-01-09
running set_trace() multiple times displays stdlib bdb.py


bukzor:
Reproduction steps:
1. Save below script as foobar.py
2. `python foobar.py foo bar`
3. You'll break at function `foo`
4. Defined a breakpoint at `foobar.py:11`  (the first line of `main()`: `for arg in argv:`)
5. c
6. You'll be brought to line 11
7. c
8. You'll be brought to `/usr/lib/python2.6/bdb.py` at `_ste_stopinfo()`, and also see that you now have two breakpoints at `foobar.py:11`
9. Pressing `s` fourteen times eventually brings you to the correct breakpoint in `bar()`

The expected behavior is to not display bdb.py, and not create duplicate breakpoints.

Obviously there are other ways to accomplish the desired debugging session, but in a large, complex code base, it is sometimes extremely convenient to have multiple set_trace lines, which may or may not cause set_trace to be called twice.

If you can tell me how to induce this error automatically, I can write a unit test for it, at minimum, and likely produce a patch for the bug as well.

```
# Script: foobar.py
def foo():
    import pudb; pudb.set_trace()
    print 'foo'

def bar():
    import pudb; pudb.set_trace()
    print 'bar'

def main(argv):
    for arg in argv:
        if arg == 'foo':
            foo()
        elif arg == 'bar':
            bar()

import sys
main(sys.argv)
```

PR #610, opened 2023-08-30
Do not raise when an instance already exists (fixes issue #607)

souliane:


pudb/debugger.py
```diff
@@ -186,8 +186,6 @@ class Debugger(bdb.Bdb):
     def __init__(self, stdin=None, stdout=None, term_size=None, steal_output=False,
                  _continue_at_start=False, **kwargs):
 
-        if Debugger._current_debugger:
-            raise ValueError("a Debugger instance already exists")
```
inducer, line 190: The problem with this is that various parts of the code assume that only one of these objects ever exists, e.g. here:

https://github.com/inducer/pudb/blob/b9c19487fa2d85c3da2eabcfa79f6732df4fe79d/pudb/__init__.py#L74-L90

souliane: I see... can you think of way to fix the issue without messing with those parts of the code? Maybe we can destroy the pudb instance when `c` is pressed?

asmeurer:
This seems like the wrong fix. If the debugger is already running then further breakpoints should just break the debugger. If that doesn't happen we should try to investigate why. Having multiple debugger classes implies recursive debugging, which sounds really complicated and probably not worth supporting.

hwalinga:
@asmeurer Hello

I did an analysis of this. If you use `pytest --pdbcls pudb.debugger:Debugger --pdb --capture=no test.py`, pytest instantiates a new Debugger class (1) whenever a new breakpoint is encountered (!! with PYTHONBREAKPOINT not set !!) and (2) on post-mortem analysis. 

Making Debugger a singleton using metaclasses should solve this. HOWEVER, an old bug from #52 resurfaced then. I tried to fix this in #667 

What is more, pytest-pudb also has exhibits the bug in #52 when using below code:

``` 
def test_foo():
    breakpoint()
    raise Exception
```

Issue #542, opened 2022-08-19
Hide the header


sisrfeng:
How can I hide the header?
![image](https://user-images.githubusercontent.com/53520949/185631880-0059579d-7cf9-42fe-ae43-c97937912181.png)

inducer:
There's not currently a way. I'd be up for looking at a PR.

PR #614, opened 2023-09-29
Added setting to hide the header

ThomasJRyan:
Now you see it
image

image
image

image

Now you don't

In reference to this issue 
https://github.com/inducer/pudb/issues/542

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -2841,7 +2841,7 @@ Error with jump. Note that jumping only works on the topmost stack frame.
                 (None, " "),
                 ("header warning", "[POST-MORTEM MODE]")
                 ])
-            self.update_header(override=True)
+            self.show_header()
             CONFIG["hide_header"] = False
 
         elif exc_tuple is not None:
@@ -2849,7 +2849,7 @@ Error with jump. Note that jumping only works on the topmost stack frame.
                 (None, " "),
                 ("header warning", "[PROCESSING EXCEPTION - hit 'e' to examine]")
                 ])
-            self.update_header(override=True)
+            self.show_header()
             CONFIG["hide_header"] = False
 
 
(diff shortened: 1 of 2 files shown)


pudb/debugger.py
```diff
@@ -2953,6 +2962,11 @@ def make_frame_ui(i, frame_lineno):
 
     def update_cmdline_win(self):
         self.set_cmdline_state(not CONFIG["hide_cmdline_win"])
+        
+    def update_header(self, override=None):
+        if override is not None:
+            self.top._w.header = self.header if override else None
+        self.top._w.header = self.header if not CONFIG["hide_header"] else None
```
inducer, line 2969: Is there an `else` missing here? Otherwise, it would seem the `override` logic would be ineffective. (Why is it needed anyhow?)

ThomasJRyan: Actually, there's a return missing here between these lines. Good catch. I knew something was off with it.

The `override` is just a glorified `show_header` function that I squished into a single function. In hindsight there's no reason to not just have a separate function for it. The purpose was for showing exceptions since the warning for it occurs in the header and it needs to be displayed even if the header is hidden.

I'll break it up into two clearer functions and you can see how you feel about it

pudb/debugger.py
```diff
@@ -2838,11 +2841,16 @@ def interaction(self, exc_tuple, show_exc_dialog=True):
...
+            self.show_header()
+            CONFIG["hide_header"] = False
+
         elif exc_tuple is not None:
             caption.extend([
                 (None, " "),
                 ("header warning", "[PROCESSING EXCEPTION - hit 'e' to examine]")
                 ])
+            self.show_header()
+            CONFIG["hide_header"] = False
```
inducer, line 2853: This is a little weird; at this point the on-disk configuration will start disagreeing from the in-memory one. I'm guessing if one opens the preferences dialog, this changed config will then be persisted to disk, but not otherwise, which may be confusing.

Overall, I find this code a bit too stateful. I would like it better if there were simply one place that decides whether the header is shown (if configured so, or if an there's an exception), and then all the places that may cause state changes affecting this decision call that central place.

PR #616, opened 2023-10-07
Exclude "test" from installed packages

mgorny:
Fix `setup.py` not to install the `test` directory as a top-level package, i.e. as:

    /usr/lib/python3.11/site-packages/test


inducer:
Thanks for the heads-up! (That's embarrassing.)

I've decided to pursue another approach, see #619.

PR #620, opened 2023-11-12
Add restart hotkey

s-redko:
It's useful to be able to restart debugging session quickly without using `q` command and selecting `<Restart>` from the UI.
It also remove the need to have `Prompt before quitting` enabled so `pudb` can be closed faster as well.

### DIFF
--- a/pudb/__init__.py
+++ b/pudb/__init__.py
@@ -172,6 +172,10 @@ def runscript(mainpyfile, args=None, pre_run="", steal_output=False,
             import urwid
             pre_run_edit = urwid.Edit("", pre_run)
 
+            if dbg.restart_requested():
+                dbg.clear_restart()
+                break
+
             if not load_config()["prompt_on_quit"]:
                 return
 
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -196,6 +196,7 @@ class Debugger(bdb.Bdb):
         self.ui = DebuggerUI(self, stdin=stdin, stdout=stdout, term_size=term_size)
         self.steal_output = steal_output
         self._continue_at_start__setting = _continue_at_start
+        self.__restart = False
 
         self.setup_state()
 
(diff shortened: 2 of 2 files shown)


inducer:
How about adding an "r" binding in the quit dialog, here:

https://github.com/inducer/pudb/blob/8c8f9d4a25e226e77a18d4d938eecf2f969e9b32/pudb/__init__.py#L197

Then you could hit "qr" in quick succession and get the same result?

s-redko:
> How about adding an "r" binding in the quit dialog, here:

I didn't know about that. However it still two keystrokes instead of just one.
I'll try to explain my motivation in more details.
When I try to find out where the error is coming from it usually requires multiple restarts while playing around with breakpoints. So it is very convenient to have single hotkey restart option.
Once bug is identified and I'm trying to fix it or add more debugging code I'm also following `edit -> restart -> edit` cycle.
Lastly it's very handy to be able to just hit `q` and close `pudb` immediately (`Prompt before exit` disabled)

I've happy to change `r` hotkey to something else. I just felt natural to use `r` for restart. Especially since `r` and `f` trigger the same action.

I also would like to add code switch focus to the `code` window inside `restart` function. I found that I usually call it from `Breakpoints` window and then have to switch to `Code` window. to hit `c`.

I do understand that it's just my workflow and might not apply for the majority of the users so I follow your suggestions.
Cheers

asmeurer:
I like the `qr` idea. I'm a little hesitant to mapping a single key to restart because it would be too easy to press accidentally. Unless of course it pops up a confirmation dialog, but that sort of defeats the purpose and would be no different from `qr`.

asmeurer:
By the way, we should indicate the Q in "quit" is a shortcut in the same way we do for the "variables" and "breakpoints" headers.

s-redko:
Should I revert my changes and add mapping of `r` shortcut to `quit` dialog box in thin PR? Or should I close this one and create a new one?

inducer:
You can just push to this branch.

s-redko:
Done. I've been using it for a while now locally. I'm happy with this solution.

pudb/__init__.py
```diff
@@ -196,6 +196,7 @@ def runscript(mainpyfile, args=None, pre_run="", steal_output=False,
                 extra_bindings=[
                     ("q", "quit"),
                     ("esc", "examine"),
+                    ("r", "restart"),
```
inducer, line 199: Please add "r" to the text in the text box, to document the feature.

Issue #628, opened 2024-01-06
Allow setting term size via env var


max-arnold:
For remote debugging it would be useful to be able to set `PUDB_TERM_SIZE=80x20` to override the default guess.

This would nicely complement `PYTHONBREAKPOINT`, `PUDB_RDB_HOST`, and `PUDB_RDB_PORT`

max-arnold:
Merged

PR #629, opened 2024-01-06
Allow defining term_size via PUDB_TERM_SIZE env var

max-arnold:
Implement #628

### DIFF
--- a/doc/starting.rst
+++ b/doc/starting.rst
@@ -80,6 +80,16 @@ The terminal size can be defined via the environment variable as well::
 
     export PUDB_TERM_SIZE=80x24
 
+The following precedence (from highest to lowest) is used to determine
+the terminal size:
+
+#. ``term_size`` keyword argument
+#. ``PUDB_TERM_SIZE`` environment variable
+#. Size of the terminal in which the debugged program is running
+   (as returned by ``os.get_terminal_size()``)
+#. Default fallback value of ``(80, 20)`` if one of the previous steps
+   fails with an exception
+
 At this point, the debugger will look for a free port and wait for a telnet
 connection::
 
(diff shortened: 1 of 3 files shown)


doc/starting.rst
```diff
@@ -76,6 +76,10 @@ enabling you to connect via ``telnet``, use the following code::
     from pudb.remote import set_trace
     set_trace(term_size=(80, 24))
 
+The terminal size can be defined via the environment variable as well::
+
+    export PUDB_TERM_SIZE=80x24
+
```
inducer, line 82: Please document which takes precedence in the various cases.

max-arnold: Done!

In your opinion, does it make sense to support something like `PUDB_RDB_REVERSE` for the sake of completeness?

inducer: What would the use case be? I.e. in what situation would you be unable to initiate "reverse" debugging in the existing ways?

max-arnold: The main use-case is using the generic `breakpoint()` to invoke the debugger. Then the debugger parameters have to be defined as env vars (`PYTHONBREAKPOINT`, etc)

inducer: Not sure I'm sold. If at all, let's do that in a separate PR.

pudb/remote.py
```diff
@@ -139,8 +141,10 @@ def __init__(
 
         if term_size is None:
             try:
-                s = os.get_terminal_size()
-                term_size = (s.columns, s.lines)
+                term_size = tuple(map(int, PUDB_TERM_SIZE.split("x")))
+                if len(term_size) != 2:
```
inducer, line 145: Didn't spot this yesterday: I don't like that this is silently ignored if the format isn't right. It'd be better to raise. (Also above.)

max-arnold: Done. Now if the `PDB_TERM_SIZE` value is invalid, one of the following exceptions will be raised:

- `ValueError: invalid literal for int() with base 10`
- `ValueError: PUDB_TERM_SIZE should have two dimensions`

inducer:
LGTM, thanks!

max-arnold:
Any chances to see it released soon?

PR #631, opened 2024-01-16
Add PUDB_RDB_REVERSE var to control reverse mode

max-arnold:
A follow-up PR to #629. 

A user can enable reverse debugging mode by setting the `PUDB_RDB_REVERSE` env var. This is useful when the debugger is invoked via `breakpoint()` ([PEP 553](https://peps.python.org/pep-0553/)) and there are multiple breakpoints


inducer:
Thanks!

inducer:
https://pypi.org/project/pudb/2024.1/

PR #638, opened 2024-03-05
fix: macOS PUDB_TTY "I/O operation on closed file."

bukzor:
Don't close a file that's in use.
Apparently linux is okay with this, but macos isn't?

Otherwise, I get this error:

```
  File ".../site-packages/pudb/debugger.py", line 2542, in call_with_ui
    return f(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^
  File ".../site-packages/pudb/debugger.py", line 2865, in interaction
    self.event_loop()
  File ".../site-packages/pudb/debugger.py", line 2821, in event_loop
    self.screen.draw_screen(self.size, canvas)
  File ".../site-packages/urwid/display/_raw_display_base.py", line 582, in draw_screen
    self._setup_G1()
  File ".../site-packages/urwid/display/_raw_display_base.py", line 524, in _setup_G1
    self.write(escape.DESIGNATE_G1_SPECIAL)
  File ".../site-packages/urwid/display/_raw_display_base.py", line 220, in write
    self._term_output_file.write(data)
ValueError: I/O operation on closed file.
```

### DIFF
--- a/pudb/__init__.py
+++ b/pudb/__init__.py
@@ -80,7 +80,6 @@ def _get_debugger(**kwargs):
             kwargs.setdefault("stdin", tty_file)
             kwargs.setdefault("stdout", tty_file)
             kwargs.setdefault("term_size", term_size)
-            tty_file.close()
 
         from pudb.debugger import Debugger
         dbg = Debugger(**kwargs)


pudb/__init__.py
```diff
@@ -80,7 +80,6 @@ def _get_debugger(**kwargs):
             kwargs.setdefault("stdin", tty_file)
             kwargs.setdefault("stdout", tty_file)
             kwargs.setdefault("term_size", term_size)
-            tty_file.close()
```
inducer, line 83: I agree that the current state is incorrect.

Please either add a comment that we're purposefully leaking the FD, or make a way for `Debugger` to close the FD when done.

bukzor: I know this isn't what you mean but: The file closes when the process ends. We're doing one-time initialization of resources which will be used in the indefinite future, so it seems valid to let reference counting take care of it, to me.

I'll look and see whether I can find a better solution though.

inducer:
Superseded by #656.

Issue #598, opened 2023-04-10
“I/O operation on closed file” when try with "Debugging from a separate terminal"


guoyejun:
Hi, I met “I/O operation on closed file” when try with "Debugging from a separate terminal" at https://documen.tician.de/pudb/starting.html#debugging-from-a-separate-terminal.  

I'm using conda with below versions:
```
$ python --version
Python 3.9.7
$ pudb --version    
pudb v2022.1.3
$ which pudb
/home/testit/miniconda3/envs/trypudb/bin/pudb
```

my python code:
```
$ cat try.py
from datetime import datetime
import os
import sys

from pudb.remote import set_trace
set_trace()

if __name__ == "__main__":
    print(' '.join(sys.argv))
    begin = datetime.now()
    print("begin process", os.getpid(), "at", begin.strftime('%Y-%m-%d %H:%M:%S'))
```

On the first terminal, I tried with:
```
$ tty
/dev/pts/3
$ perl -MPOSIX -e pause
```

On the second terminal, I met the below issue.  Should I run with 'root'? thanks.
```
$ PUDB_TTY=/dev/pts/3 pudb try.py
Traceback (most recent call last):
  File "/home/testit/miniconda3/envs/trypudb/lib/python3.9/site-packages/pudb/debugger.py", line 2434, in call_with_ui
    return f(*args, **kwargs)
  File "/home/testit/miniconda3/envs/trypudb/lib/python3.9/site-packages/pudb/debugger.py", line 2740, in interaction
    self.event_loop()
  File "/home/testit/miniconda3/envs/trypudb/lib/python3.9/site-packages/pudb/debugger.py", line 2696, in event_loop
    self.screen.draw_screen(self.size, canvas)
  File "/home/testit/miniconda3/envs/trypudb/lib/python3.9/site-packages/urwid/raw_display.py", line 708, in draw_screen
    self._setup_G1()
  File "/home/testit/miniconda3/envs/trypudb/lib/python3.9/site-packages/urwid/raw_display.py", line 687, in _setup_G1
    self.write(escape.DESIGNATE_G1_SPECIAL)
  File "/home/testit/miniconda3/envs/trypudb/lib/python3.9/site-packages/urwid/raw_display.py", line 286, in write
    self._term_output_file.write(data)
ValueError: I/O operation on closed file.

```

axman6:
I’m also having this issue, which is problematic because when the app I’m debugging prints anything, the pudb UI disappears until I terminate the app.

[...]

PR #656, opened 2024-07-14
Fix separate-terminal debugging

MatrixManAtYrService:
This PR reverts part of https://github.com/inducer/pudb/pull/534, which appears to have broken the ability to debug in a separate terminal (see comment there for more).

Presumably there was a reason that we wanted to close this file, but I'm not sure what that reason was. Leaving it open restored this functionality for me and nothing else bad appears to be happening 🤷 .

Maybe @inducer can suggest a better place for closing the file so that the needs of https://github.com/inducer/pudb/pull/534 are met while also not breaking separate-terminal debugging?

### DIFF
--- a/pudb/__init__.py
+++ b/pudb/__init__.py
@@ -83,6 +83,7 @@ def _get_debugger(**kwargs):
             kwargs.setdefault("stdin", tty_file)
             kwargs.setdefault("stdout", tty_file)
             kwargs.setdefault("term_size", term_size)
+            kwargs.setdefault("tty_file", tty_file)
 
         from pudb.debugger import Debugger
         dbg = Debugger(**kwargs)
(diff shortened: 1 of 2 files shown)


MatrixManAtYrService:
Fixes https://github.com/inducer/pudb/issues/598

MatrixManAtYrService:
I just realized there's already one for this: https://github.com/inducer/pudb/pull/638

Sorry for the noise.

inducer:
This looks good to me; thanks for working on it. I'll roll a release to get this out into the world.

MatrixManAtYrService:
Sorry for being indecisive.  I had a problem which I thought might be introduced by this change, so I got paranoid and marked it as a draft with the idea that I'd use it for a while before calling it good.  As it turns out, that problem was entirely separate.

I used it for a while, but I forgot to circle back and un-draft it.  I think this is indeed an improvement.

inducer:
Thanks!

inducer:
https://pypi.org/project/pudb/2024.1.2/

PR #659, opened 2024-08-29
Fix startup when no write permissions

fffergal:
Making the parent directories of the config file could cause an exception if the process doesn't have write permissions. This is common with remote debugging in containers. Catch exceptions, log, and continue, same as other parts of the settings that cannot raise exceptions. The next stage of saving the config file will also fail, but this is already caught.

### DIFF
--- a/pudb/settings.py
+++ b/pudb/settings.py
@@ -49,7 +49,10 @@ def get_save_config_path():
         return None
 
     path = os.path.join(XDG_CONFIG_HOME, XDG_CONF_RESOURCE)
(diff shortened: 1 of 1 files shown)


fffergal:
Was waiting for checks to run before making ready, but I don't think they are going to run without intervention so making ready now. Anything I can do to make this easier to merge and release, just let me know! Thank you.

inducer:
Thanks!

inducer:
Sounds good. Happy to consider further PRs. And thanks for contributing!

PR #664, opened 2024-10-15
Fix compatibility with Python 3.13

inducer:
This fixes a few issues I've noticed with Python 3.13, which seems to contain quite a few changes to `bdb`. This seems to work for me, but I'm not necessarily confident in these changes. See the commit messages for some explanation.

I would love some testing from others before I release this.

@asmeurer @mvanderkamp Would you be willing to give this a look/test?

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -456,7 +456,7 @@ class Debugger(bdb.Bdb):
         if self._waiting_for_mainpyfile(frame):
             return
 
-        if "__exc_tuple__" not in frame.f_locals:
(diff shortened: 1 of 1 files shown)


inducer:
Ah well, it's probably better to have this than not. I'll merge and roll a release.

asmeurer:
The code here seems to be related to the work I did on unbreakpointing a set_trace line. I tested it in 3.9, 3.12, and 3.13 and it still seems to work.

inducer:
Thanks for taking a look!

PR #667, opened 2024-11-29
Make the Debugger a singleton and do not reset the stopframe to None during postmortem.

### LINKED ISSUE #607: "Debugger instance already exists" when used in pytest

hwalinga:
This fixes #607. Making the Debugger a singleton should be enough to fix #607. HOWEVER, pytest def postmortem() calls .reset() which resets the stopframe. Somehow, this causes the debugger to stop somewhere inside internal debugger source code, instead of in the code the postmortem should be done.

So, this also includes an override of the .reset() method and detects when the debugger is already running and does not reset the stopframe. See also #67.

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -183,14 +183,20 @@ OTHER DEALINGS IN THE SOFTWARE.
 
 # {{{ debugger interface
 
-class Debugger(bdb.Bdb):
+class Singleton(type):
+    _instance = None
+
+    def __call__(cls, *args, **kwargs):
+        if not cls._instance:
+            cls._instance = super(Singleton, cls).__call__(*args, **kwargs)
+        return cls._instance
+
+
+class Debugger(bdb.Bdb, metaclass=Singleton):
     _current_debugger = []
 
     def __init__(self, stdin=None, stdout=None, term_size=None, steal_output=False,
                  _continue_at_start=False, tty_file=None, **kwargs):
-
-        if Debugger._current_debugger:
-            raise ValueError("a Debugger instance already exists")
 
         # Pass remaining kwargs to python debugger framework
         bdb.Bdb.__init__(self, **kwargs)
@@ -213,6 +219,30 @@ class Debugger(bdb.Bdb):
 
         # Okay, now we have a debugger
         self._current_debugger.append(self)
+
+    def reset(self):
+        """Set values of attributes as ready to start debugging.
+
+        Override from Bdb.reset()
+
+        When pytest starts a postmortem analysis, but the debugger is already active,
+        calling .reset() in src/_pytest/debugging.py::post_mortem
+        causes the self.stopframe to be set to None.
+        This pauses the debugger somewhere in the source code of the debugger. See #67
+
+        We detect using _current_debugger that this is the case and do not set the
+        stopframe to None then.
+
+        Related #607, #52
+        """
+        import linecache
+        linecache.checkcache()
+        self.botframe = None
+        if not self._current_debugger:
+            self.stopframe = None
+        self.returnframe = None
+        self.quitting = False
+        self.stoplineno = 0
 
     def __del__(self):
         # according to https://stackoverflow.com/a/1481512/1054322, the garbage


hwalinga:
Reproduce issue. 

Install pytest and pudb.

`test.py`:

```py
def test_foo():
    breakpoint()
    raise Exception
```

```sh
pytest --pdbcls pudb.debugger:Debugger --pdb --capture=no test.py
```

NB. The stopframe issue (#52) can also be observed with pytest-pudb using

```sh
pytest --pudb test.py
```

pudb/debugger.py
```diff
@@ -214,6 +223,32 @@ def __init__(self, stdin=None, stdout=None, term_size=None, steal_output=False,
...
+        import linecache
+        linecache.checkcache()
+        self.botframe = None
+        if not self._pytest_postmortem:
+            self.stopframe = None
+        else:
+            self._pytest_postmortem = False
+        self.returnframe = None
+        self.quitting = False
+        self.stoplineno = 0
```
inducer, line 250: Would it be possible to call into `Bdb.reset` for some of this? We're already copy-pasting too much of Bdb's guts into pudb... (If not, add a comment why not.)

pudb/debugger.py
```diff
@@ -214,6 +223,32 @@ def __init__(self, stdin=None, stdout=None, term_size=None, steal_output=False,
         # Okay, now we have a debugger
         self._current_debugger.append(self)
```
inducer, line 224: Could we get rid of `_current_debugger` in favor of the singleton mechanism?

pudb/debugger.py
```diff
@@ -214,6 +223,32 @@ def __init__(self, stdin=None, stdout=None, term_size=None, steal_output=False,
...
 
+    def reset(self):
+        """Set values of attributes as ready to start debugging.
+
+        Override from Bdb.reset()
+
+        When pytest starts a postmortem analysis, but the debugger is already active,
+        calling .reset() in src/_pytest/debugging.py::post_mortem
```
inducer, line 232: Add Github link to specific version?

inducer (commented):
Thanks. Some comments below.

pudb/debugger.py
hwalinga: That is logic in place for `_get_debugger`. This has to remain. That is all fine.

Trying to instantiate a debugger again (and thus previously hitting the ValueError) is what pytest is doing. Thus my code only works around that specific pytest problem. All logic around `_get_debugger` is fine, but pytest is just not calling it.

pudb/debugger.py
hwalinga: Yes

pudb/debugger.py
hwalinga: When calling Bdb.reset the debugger stops around there. `self.stopframe = None` is the offending line. Calling Bdb.reset is a no-go. It is the exact same problem for #67 where also code is copies from Bdb in here. The problem now is that pytest is calling .reset, so that is why the override is needed and to check for the pytest postmortem state. I can add a link to the source of pytest as well, but I believe the docstring of this method explain already enough.

Do you think there needs to be more information in the docstring of this method to explain why it needs to be overriden?

hwalinga:
@inducer Added more comments.

inducer:
Unsubscribing... @-mention or request review once it's ready for a look or needs attention.

asmeurer:
Does this affect recursive debugging? I know that's not something that we should always expect to work, but I think it is something that has worked at least in limited cases.

hwalinga:
@asmeurer What does recursive debugging mean here? Do you mean debugging a recursive function? That would not be affected.

PR #674, opened 2025-01-18
debugger: leave output screen with single key press

gsigh:
This is a cleaned up submission of the PR 673 essence. Implements transparent support for platform dependent single key press without a prompt, with Python core and without external dependencies. Supports all major desktop systems where you expect to run the debugger, falls back to the most portable backwards compatible behaviour on platforms that are known to not work with the single key approach. Does not involve config items, needs no documentation update. The implementation is considered both readable and maintainable, supporting more platforms is straight forward as they get identified. Thank you for the feedback which helped very much in getting to that state.

### DIFF
--- a/pudb/lowlevel.py
+++ b/pudb/lowlevel.py
@@ -334,7 +334,7 @@ class ConsoleSingleKeyReader(object):
             # special keys (function keys, cursor, keypad) require
             # another call when the first returned '\0' or '\xe0'.
             c = msvcrt.getch()
-            if c in ('\x00', '\xe0'):
+            if c in ("\x00", "\xe0"):
                 c = msvcrt.getch()
             return c
 
--- a/pudb/lowlevel.py
+++ b/pudb/lowlevel.py
@@ -293,21 +293,18 @@ def decode_lines(lines):
 # Is used in situations where urwid is disabled and curses calls are
 # not available.
 
-_nbc_use_input = True
-_nbc_use_getch = False
-_nbc_use_select = False
+( _NBC_IMPL_INPUT, _NBC_IMPL_GETCH, _NBC_IMPL_SELECT, ) = range(3)
+_nbc_impl = _NBC_IMPL_INPUT
 if sys.platform in ("emscripten", "wasi"):
     pass
 elif sys.platform in ("win32",):
     import msvcrt
-    _nbc_use_input = False
-    _nbc_use_getch = True
+    _nbc_impl = _NBC_IMPL_GETCH
 else:
     import select
     import termios
     import tty
-    _nbc_use_input = False
-    _nbc_use_select = True
+    _nbc_impl = _NBC_IMPL_SELECT
 
 
 class NonBufferedConsole(object):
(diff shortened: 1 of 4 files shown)


pudb/lowlevel.py
```diff
@@ -283,4 +283,73 @@ def decode_lines(lines):
...
+# Local platform dependent helper to get single key presses from a
+# terminal in unbuffered mode. Eliminates the necessity to press ENTER
+# before other input also becomes available. Also avoids the accumulation
+# of prompts on the screen as was the case with Python's input() call.
+# Is used in situations where urwid is disabled and curses calls are
+# not available.
+
+_nbc_use_input = True
+_nbc_use_getch = False
+_nbc_use_select = False
```
inducer, line 298: Could you make this one enum instead of multiple booleans?

pudb/lowlevel.py
```diff
@@ -283,4 +283,73 @@ def decode_lines(lines):
...
 
+
+# {{{ non-buffered console
+
+# Local platform dependent helper to get single key presses from a
+# terminal in unbuffered mode. Eliminates the necessity to press ENTER
+# before other input also becomes available. Also avoids the accumulation
+# of prompts on the screen as was the case with Python's input() call.
+# Is used in situations where urwid is disabled and curses calls are
+# not available.
```
inducer, line 294: Make this the class docstring?

pudb/lowlevel.py
```diff
@@ -283,4 +283,73 @@ def decode_lines(lines):
...
+    _nbc_use_getch = True
+else:
+    import select
+    import termios
+    import tty
+    _nbc_use_input = False
+    _nbc_use_select = True
+
+
+class NonBufferedConsole(object):
```
inducer, line 313:
```suggestion
class ConsoleSingleKeyReader:
```
?

pudb/lowlevel.py
```diff
@@ -283,4 +283,73 @@ def decode_lines(lines):
...
+
+    def __init__(self):
+        pass
+
```
inducer, line 317:
```suggestion
```

pudb/lowlevel.py
```diff
@@ -283,4 +283,73 @@ def decode_lines(lines):
...
+        return self
+
+    def __exit__(self, type, value, traceback):
+        if _nbc_use_select:
+            termios.tcsetattr(sys.stdin, termios.TCSADRAIN, self.prev_settings)
+
+    def get_data(self):
+        if _nbc_use_getch:
+            c = msvcrt.getch()
+            if c in ('\x00', '\xe0'):
```
inducer, line 331: Could you add a comment on what these mean?

pudb/lowlevel.py
```diff
@@ -283,4 +283,73 @@ def decode_lines(lines):
...
+                c = msvcrt.getch()
+            return c
+
+        if _nbc_use_select:
+            rset, _, _ = select.select([sys.stdin], [], [], None)
+            if sys.stdin in rset:
+                return sys.stdin.read(1)
+            return None
```
inducer, line 339: Could you explain the fall-through to `return None`? When would `sys.stdin` not be in `rset`? I would say the `if` could be an `assert`.

pudb/lowlevel.py
```diff
@@ -283,4 +283,73 @@ def decode_lines(lines):
...
+        if _nbc_use_select:
+            self.prev_settings = termios.tcgetattr(sys.stdin)
+            tty.setcbreak(sys.stdin.fileno())
+        return self
+
+    def __exit__(self, type, value, traceback):
+        if _nbc_use_select:
+            termios.tcsetattr(sys.stdin, termios.TCSADRAIN, self.prev_settings)
+
+    def get_data(self):
```
inducer, line 328:
```suggestion
    def get_single_key(self):
```

inducer (commented):
Thanks! Some comments/questions below.

gsigh:
ACK on all your feedback items. Have quickly pushed 515ecc86f735..bc27f149bd60 and will later look at lint and ruff, and test some more.

Got ideas or preferences for the enum names? Drop the "impl" in the enum values? Keep _KEYREAD_GETCH et al for the available choices, use _keyread_impl for the variable that's dispatched on?

inducer:
Pushed some fixes. How does this look to you?

inducer:
Could you comment on your plan? I appreciate the thorough testing, but I also would like to move on from this. If you are OK with this, my preference would be to remove your `nonbufferedconsole.py` test script and merge this. What's your take?

gsigh:
Have pushed the test script this morning to run more tests, as I mentioned earlier ("will test more"). Have verified the operation of the approach on Windows and Cygwin again, while operation on Linux was known and Mac was reasonably assumed. Did not intend to confuse you after you spent so much time in the thorough review, which I appreciate very much.

Have pushed the cleaned up result now in the afternoon, which you can pick up in my opinion. Have taken in all of your feedback. When you are happy you could merge --ff-only my v3 branch, or cherry-pick the single commit. Or tell if you want me to submit a PR for v3. Just don't take v2 which is dirty, has expectedly been while review is ongoing. (And no I don't believe in force pushing, or squashing during merges.)

You can probably tell how I'm not familiar with the Github flow, and neither am fluent with most Python idioms or tooling. But am trying as hard as I can to cooperate, and provide something that is acceptable to you, and usable in the project. Hope that v3 matches your expectation.

inducer:
I've merged this manually. Thanks for your contribution.

PR #676, opened 2025-02-02
Update starting.rst

martin-kokos:
Add persistent pytest config method

### DIFF
--- a/doc/starting.rst
+++ b/doc/starting.rst
@@ -179,5 +179,7 @@ like so::
 
     $ pytest --pdbcls pudb.debugger:Debugger --pdb --capture=no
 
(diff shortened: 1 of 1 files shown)


doc/starting.rst
```diff
@@ -179,5 +179,7 @@ like so::
 
     $ pytest --pdbcls pudb.debugger:Debugger --pdb --capture=no
 
+More persistantly, it's possible to add the option to the ``adopts =`` settings in one of pytest's `configuration files <https://docs.pytest.org/en/stable/reference/customize.html#configuration-file-formats>`_.
```
inducer, line 182:
```suggestion
To make this persistent, it is possible to add the option to the ``adopts =`` settings in one of pytest's `configuration files <https://docs.pytest.org/en/stable/reference/customize.html#configuration-file-formats>`_.
```

PR #688, opened 2025-06-09
Fix set_trace(paused=False)

asmeurer:
Previously this gave an exception like

```
>>> pudb._get_debugger().set_trace(paused=False)
  File "<stdin>", line 1, in <module>
  File "/Users/aaronmeurer/miniconda3/lib/python3.11/site-packages/pudb/debugger.py", line 291, in set_trace
    self.set_continue()
  File "/Users/aaronmeurer/miniconda3/lib/python3.11/bdb.py", line 344, in set_continue
    self._set_stopinfo(self.botframe, None, -1)
                       ^^^^^^^^^^^^^
AttributeError: 'Debugger' object has no attribute 'botframe'
```

The fix is to always set the botframe variable, even when paused=False, because after commit 2b08f08910da00e58df736c719f11ae75bd5333c, the bdb set_continue method is used, which requires botframe to be set.

### DIFF
--- a/pudb/debugger.py
+++ b/pudb/debugger.py
@@ -31,16 +31,29 @@ import bdb
 import gc
 import os
 import sys
+from abc import ABC, abstractmethod
 from collections import deque
 from functools import partial
 from itertools import count
-from types import TracebackType
-from typing import ClassVar
+from os.path import splitext
+from types import FrameType, ModuleType, TracebackType
+from typing import TYPE_CHECKING, Any, ClassVar, TextIO, TypeVar, cast, final
 
 import urwid
+from typing_extensions import ParamSpec, override
 
 from pudb.lowlevel import ConsoleSingleKeyReader, decode_lines, ui_log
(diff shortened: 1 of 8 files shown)


pudb/debugger.py
```diff
@@ -262,24 +262,22 @@ def set_trace(self, frame=None, as_breakpoint=None, paused=True):
         # See pudb issue #52. If this works well enough we should upstream to
         # stdlib bdb.py.
         # self.reset()
-
-        if paused:
```
asmeurer, line 266: FYI the only change here is removing this line (this is clearer if you look at https://github.com/inducer/pudb/pull/688/files?w=1)

inducer:
Well, this has sent me down quite the rabbit hole, including the discovery of a typeshed bug: https://github.com/python/typeshed/pull/14502. :slightly_smiling_face:

asmeurer:
Wow I don't know if I could have figured all that out.

PR #691, opened 2025-07-04
Expose `post_mortem` (and `pm`) in `pudb.remote`

### LINKED ISSUE #690: Expose post_mortem() feature in RemoteDebugger.
**Is your feature request related to a problem? Please describe.**

While `pudb.remote` exposes set_trace, it doesn’t expose the very useful post_mortem() or pm() functions.
I ended up finding the right incantation after getting weird errors about _tty_file or boolean not callable.

```py
             try:
                 do_something()
             except Exception:
                    import sys
                    from pudb.remote import debugger
                    dbg = debugger(term_size=(150, 50), port=4444)
                    dbg.reset()
                    dbg.interaction(None, sys.exc_info())
```

**Describe the solution you'd like**
expose post_mortem at a top level function in `pudb.remote` would probably be ideal. With code similar to what’s above.

**Describe alternatives you've considered**

I though initially that having `pudb.remote.debugger().post_mortem()` would be fine, but it would be weird considering post_mortem is a bool. that is used for book keeping inside `Debugger` the parent class of `RemoteDebugger`.

**Additional context**
None

tshirtman:
Closes #690

### DIFF
--- a/.basedpyright/baseline.json
+++ b/.basedpyright/baseline.json
@@ -186,56 +186,6 @@
                 }
             }
         ],
-        "./doc/conf.py": [
-            {
-                "code": "reportAny",
-                "range": {
-                    "startColumn": 27,
-                    "endColumn": 31,
-                    "lineCount": 1
-                }
-            },
-            {
-                "code": "reportAny",
-                "range": {
-                    "startColumn": 17,
-                    "endColumn": 26,
-                    "lineCount": 1
-                }
-            },
-            {
-                "code": "reportAny",
-                "range": {
-                    "startColumn": 17,
-                    "endColumn": 28,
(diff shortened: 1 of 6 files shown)


tshirtman:
Basedpyright is happy locally now, let's see how it is in CI.

inducer:
> Ruff suggests using `|` instead of `Union` and `tuple[…]` instead of `Tuple[…]`, but (according to pyproject.toml) the project supports 3.8, which doesn’t have these syntax. Guidance appreciated.

Ruff is right. Due to `from __future__ import annotations`, the annotations are not interpreted at run time.

As for the remaining unknown type with basedpyright, feel free to sweep that into the baseline with `--writebaseline`.

PR #741, opened 2026-06-05
fix: handle bool defaults in normalize_bool_inplace

Snoarlax:
I was having issues when using pudb with Docker, the config file was not writable by the pudb process due to permissions errors, causing pudb to crash. This pr adds a check to fix that error.


inducer:
Thx!

PR #745, opened 2026-07-12
Document the -s/--steal-output CLI option

rul:
Hi,

`--steal-output` was the only argument without a help string, so it appears in the `--help` output with no explanation. Add one, noting that the feature is currently unimplemented (passing `-s` raises `NotImplementedError` in `Debugger.__init__`).

Reported as [Debian bug #773262](https://bugs.debian.org/773262).


inducer:
Thx!

